PR #25, opened 2017-03-07
Add cms-image-lambda stacks

farski:


cavis:
@farski my only suggestion is for the [buildspec.yml file](https://github.com/PRX/cms-image-lambda/blob/master/buildspec.yml#L21).

Could use the `node-lambda package` command to handle the zipping for you.  And then upload it from `./build/whatever.zip`.

Also note that i've got some pending changes to the cms-node-lambda.  Shouldn't change how tests are run, though.

PR #30, opened 2017-03-22
Enable CORS on upload lambda and add error messages

gcampo88:
- [x] Adds OPTIONS method on Upload API Gateway to enable CORS
- [x] add error alarms to upload lambda; send them to Warn topic SNS


upload.prx.org/upload.prx.org.yml
```diff
@@ -83,3 +113,22 @@ Resources:
...
+      ComparisonOperator: GreaterThanThreshold
+      EvaluationPeriods: "1"
+      MetricName: Errors
+      Namespace: AWS/Lambda
+      Period: "300"
+      Statistic: Sum
+      Threshold: "0"
+      Dimensions:
+        - Name: FunctionName
+        - Value: !GetAtt UploadLambdaFunction.name
```
farski: `name` doesn't seem to be an attribute for lambda functions: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#w1ab2c19c12d594c13

upload.prx.org/upload.prx.org.yml
```diff
@@ -83,3 +113,22 @@ Resources:
       Role: !GetAtt LambdaBasicExecutionIAMRole.Arn
       Runtime: nodejs4.3
       Timeout: 5
+  UploadErrorAlarm:
+    Type: "AWS::CloudWatch::Alarm"
+    Properties:
+      ActionsEnabled: true
+      AlarmName: "Upload lambda errors"
```
farski: Might be nice to include the environment type in the name so it's more obvious which is which when there are several stacks launched

farski: Or actually, maybe it makes sense to only create this resource in production environment

upload.prx.org/upload.prx.org.yml
```diff
@@ -83,3 +113,22 @@ Resources:
       Role: !GetAtt LambdaBasicExecutionIAMRole.Arn
       Runtime: nodejs4.3
```
farski: Might as well try to use node6 since it's available and probably requires no changes

upload.prx.org/upload.prx.org.yml
```diff
@@ -83,3 +113,22 @@ Resources:
...
       Timeout: 5
+  UploadErrorAlarm:
+    Type: "AWS::CloudWatch::Alarm"
+    Properties:
+      ActionsEnabled: true
+      AlarmName: "Upload lambda errors"
+      AlarmActions:
+        - !Ref OpsDebugMessagesSNSTopicArn
```
farski: Probably we want these alarms to go to a more important SNS topic. Maybe **warn** for this one?

farski: (All those topics are made by the notifications stack)

upload.prx.org/upload.prx.org.yml
gcampo88: 👍  thanks for the link!

upload.prx.org/upload.prx.org.yml
gcampo88: 👍

upload.prx.org/upload.prx.org.yml
gcampo88: I'll throw it in there just in case.

upload.prx.org/upload.prx.org.yml
gcampo88: 👍

upload.prx.org/upload.prx.org.yml
```diff
@@ -81,5 +111,24 @@ Resources:
...
-      Runtime: nodejs4.3
+      Runtime: nodejs6.10
       Timeout: 5
+  UploadErrorAlarm:
+    Type: "AWS::CloudWatch::Alarm"
+    Properties:
+      ActionsEnabled: true
+      AlarmName: "${EnvironmentTypeAbbreviation} upload lambda errors"
+      AlarmActions:
+        - !Ref OpsWarnMessagesSNSTopicArn
```
farski: This seems to be referencing something that doesn't existing in this template

gcampo88:
@farski back to you!

upload.prx.org/upload.prx.org.yml
gcampo88: oops- should be ok now.

farski: Is that param getting passed in from Root?

gcampo88: it is now (in this PR), I believe: https://github.com/PRX/Infrastructure/pull/30/files#diff-65f4f5695da65033aa4fcf7ce0414bfbL391

root.yml
```diff
@@ -388,7 +388,7 @@ Resources:
     Condition: CreateSingletonStacks
     Properties:
       NotificationARNs:
-        - !GetAtt NotificationsStack.Outputs.OpsDebugMessagesSNSTopicArn
+        - !GetAtt NotificationsStack.Outputs.OpsWarnMessagesSNSTopicArn
```
farski: This we actually want to keep as *debug*, since only CloudFormation-related messages get sent to the Notification ARN. The topic that gets used for our other explicit resources, like CW alarms, needs to get passed in as a param

upload.prx.org/pipeline.yml
```diff
@@ -2,7 +2,7 @@ AWSTemplateFormatVersion: "2010-09-09"
 Description: >
   Creates a CodePipeline that is used to handle CI/CD for the upload lambda
 Parameters:
-  OpsDebugMessagesSNSTopicArn:
+  OpsWarnMessagesSNSTopicArn:
```
farski: It doesn't look like this ever gets used in the pipeline template. It's fine to pass in extra params, but it's unlikely a build error would ever get sent to **warn**, so in this case it may be worth taking out.

farski: Or maybe something's missing from the template, and it should be getting used?

gcampo88: I think that this was included originally because I copied over from another pipeline template that did have it -- so, happy to take it out.

root.yml
```diff
@@ -390,7 +390,7 @@ Resources:
       NotificationARNs:
         - !GetAtt NotificationsStack.Outputs.OpsDebugMessagesSNSTopicArn
       Parameters:
-        OpsDebugMessagesSNSTopicArn: !GetAtt NotificationsStack.Outputs.OpsDebugMessagesSNSTopicArn
+        OpsWarnMessagesSNSTopicArn: !GetAtt NotificationsStack.Outputs.OpsWarnMessagesSNSTopicArn
```
farski: This can't get passed in if the nested template isn't expecting it

farski:
@gcampo88 this is looking pretty good I think; just the merge conflict that has to be resolved

gcampo88:
ok @farski ! Thanks for holding my hand through parts of that!

PR #37, opened 2017-04-14
Init Castle

cavis:
Add pipelines for (new) castle.prx.tech.


cavis:
@farski okay, update `us-west-2.zip` json as well.  Should be ready to go.

PR #63, opened 2017-05-24
Lambda to update template configs when secrets files change

kookster:
invoked when an s3 file(s) in the secrets s3 bucket is updated.
uses the info from the path of the updated file to determine env and app
gets the latest version id of the secrets file
downloads and updates the template config, uploads updated file


kookster:
hmm - I should probably include the CF template for this as well...

farski:
I'm not really in favor of having this stack be responsible for the secrets S3 bucket. The way we're doing things right now, we only have a single, global bucket for configs. It remains to be seen if that will actually work out for us if we decide to deploy multi region. But for now, I think it makes sense to manage the config and secrets in a similar way, so I think probably the secrets bucket should be owned by the storage stack. If we do need region-specific configs, we would then change both similarly.

Also a little unsure about this Lambda being in the CD stack. Two main reasons for it **not** to live there: first, to this point everything in the stack is directly related to the CodePipeline (either an action in the pipeline, or something needed for the actions to work). Not to say we can't expand the scope of that, but it would be a change. Second, somewhat dependent on where the bucket lands, if there is a single global bucket, we don't want Lambdas in multiple stacks trying to watch it (not sure that can even happen cross region).

kookster:
ok, I think the lambda makes sense in CD as it is changing the same artifact, the template config zip, and is then triggering the build to start based on that for staging.

If I move the bucket to the storage stack, how do I associate the lambda to it?
Are you suggesting putting the secrets lambda in its own stack and then running it prior to the storage stack, so the storage stack can include a ref to the lambda in the s3 bucket for secrets notifications config?

kookster:
@farski can you have another look?

stacks/ecs.yml
```diff
@@ -158,25 +154,10 @@ Resources:
...
-            Version: "2012-10-17"
-            Statement:
-              - Effect: Allow
-                Action:
-                  - "s3:GetObject"
-                  - "s3:GetObjectVersion"
-                Resource:
-                  - !Sub arn:aws:s3:::${SecretsBase}-secrets/*
+        - Fn::ImportValue:
+            !Sub "${InfrastructureStorageStackName}-SecretsInstanceDecryptPolicy"
```
kookster: @farski Honestly no idea if I did this right, have to test it I guess

stacks/ecs.yml
```diff
@@ -33,11 +33,7 @@ Parameters:
     Type: String
   OpsErrorMessagesSnsTopicArn:
     Type: String
-  SecretsBase:
-    Type: String
-  SecretsKey:
-    Type: String
-  SecretsRegion:
+  InfrastructureStorageStackName:
```
kookster: I figure better to pass in this one param, already in use in the root stack, than add in any secret specific params.

secrets/secrets.yml
```diff
@@ -0,0 +1,221 @@
...
+      (.zip), inside the InfrastructureConfigBucket
+    Type: String
+Resources:
+  SecretsIAMGroup:
+    Type: "AWS::IAM::Group"
+    Properties:
+      GroupName: !Sub ${InfrastructureStorageStackName}-manage-secrets
+      Path: "/"
+      Policies:
+        - PolicyName: !Sub ${InfrastructureStorageStackName}-encrypt-secrets
```
farski: I would probably make these policy names simpler. There's no real benefit to doing the interpolation. No harm, just makes the templates a little more complicated to read.

secrets/secrets.yml
```diff
@@ -0,0 +1,221 @@
...
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action:
+              - s3:GetObject
+              - s3:GetObjectVersion
+            Resource:
+              - !Sub arn:aws:s3:::${InfrastructureStorageStackName}-secrets/*
+  SecretsKey:
```
farski: I think (kind of guessing) if we ever needed to rotate this key, we would add another one to the template, and then change what the alias points at. So it might be good to give this a name like `SecretKeyMay17` or something? idk. maybe that's dumb

secrets/secrets.yml
```diff
@@ -0,0 +1,221 @@
...
+                Resource:
+                  - !Join ["", ["arn:aws:s3:::", "Fn::ImportValue": !Sub "${InfrastructureStorageStackName}-InfrastructureConfigBucket", "/", !Ref InfrastructureConfigStagingKey]]
+                  - !Join ["", ["arn:aws:s3:::", "Fn::ImportValue": !Sub "${InfrastructureStorageStackName}-InfrastructureConfigBucket", "/", !Ref InfrastructureConfigProductionKey]]
+              - Effect: Allow
+                Action:
+                  - "s3:getObject"
+                  - "s3:getObjectVersion"
+                Resource:
+                  - !Sub arn:aws:s3:::${InfrastructureStorageStackName}-secrets/*
+        - PolicyName: CloudWatchMetricsPolicy
```
farski: Might already be a comment for this somewhere, but this Lambda doesn't need these CW permissions.

secrets/secrets.yml
kookster: for now I'm going to keep this compatible with the init script, and leave these in

farski: 👍

secrets/secrets.yml
kookster: removed

PR #70, opened 2017-06-09
Web template for apps with worker or not

kookster:
* Added param for app name to use throught template
* Added container related params for cpu, mem, port
* Added a conditional param for making worker resources or not

### DIFF
--- a/stacks/root.yml
+++ b/stacks/root.yml
@@ -77,6 +77,8 @@ Mappings:
       prefix: 10
     Play:
       prefix: 40
+    Crier:
+      prefix: 50
 Resources:
   VPCStack:
     Type: "AWS::CloudFormation::Stack"
@@ -349,7 +351,7 @@ Resources:
         PlatformALBCanonicalHostedZoneID: !GetAtt LoadBalancersStack.Outputs.PlatformALBCanonicalHostedZoneID
         PlatformALBHTTPListenerArn: !GetAtt LoadBalancersStack.Outputs.PlatformALBHTTPListenerArn
         PlatformALBHTTPSListenerArn: !GetAtt LoadBalancersStack.Outputs.PlatformALBHTTPSListenerArn
-        PlatformALBListenerPriorityPrefix: !FindInMap [ALBListenerRulePriorityMap, Publish, prefix]
+        PlatformALBListenerPriorityPrefix: !FindInMap [ALBListenerRulePriorityMap, Crier, prefix]
         ECSCluster: !GetAtt ECSClusterStack.Outputs.ECSCluster
         ECSServiceAutoscaleIAMRoleArn: !GetAtt ECSClusterStack.Outputs.ECSServiceAutoscaleIAMRoleArn
         ECSServiceIAMRole: !GetAtt ECSClusterStack.Outputs.ECSServiceIAMRole
@@ -365,8 +367,8 @@ Resources:
         SecretsBase:
             Fn::ImportValue:
               !Sub "${SecretsStackName}-SecretsBaseRegion"
-        EcrImageTag: !Ref PublishEcrImageTag
-        SecretsVersion: !Ref PublishSecretsVersion
+        EcrImageTag: !Ref CrierEcrImageTag
+        SecretsVersion: !Ref CrierSecretsVersion
         AppName: "crier"
         CreateWorker: "true"
         ContainerPort: 3000
--- a/stacks/root.yml
+++ b/stacks/root.yml
@@ -394,7 +394,7 @@ Outputs:
     Description: Domain name for play app
     Value: !GetAtt PlayStack.Outputs.HostedZoneDNSName
   CrierHostedZoneDNSName:
-    Description: Domain name for play app
+    Description: Domain name for crier app
     Value: !GetAtt CrierStack.Outputs.HostedZoneDNSName
   # CmsHostedZoneDNSName:
   #   Description: Domain name for CMS app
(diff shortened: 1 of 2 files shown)


kookster:
@farski I'm going to start trying this out in aws - any feedback?

stacks/web-application.yml
```diff
@@ -0,0 +1,257 @@
...
+      ListenerArn: !Ref PlatformALBHTTPListenerArn
+      Priority: !Join ["", [!Ref PlatformALBListenerPriorityPrefix, 03]]
+  ALBTargetGroup500Alarm:
+    Type: "AWS::CloudWatch::Alarm"
+    # Condition: CreateProductionResources
+    Properties:
+      ActionsEnabled: true
+      AlarmName: !Sub ${AppName} ALB Target Group HTTP 5XXs
+      AlarmActions:
+        - !Ref OpsErrorMessagesSnsTopicArn
```
farski: I wonder if this should be a parameter. Some apps 500'ing on the web container are a lot worse than others, and I'm trying to make sure we're being fairly deliberate about which alarms show up where.

stacks/web-application.yml
```diff
@@ -0,0 +1,257 @@
...
+    Type: "AWS::ECS::Service"
+    Properties:
+      Cluster: !Ref ECSCluster
+      DeploymentConfiguration:
+        MaximumPercent: 200
+        MinimumHealthyPercent: 50
+      DesiredCount: 1
+      Role: !Ref ECSServiceIAMRole
+      TaskDefinition: !Ref WorkerTaskDefinition
+      # TODO: set entry point to worker not web
```
farski: Is this still a TODO?

stacks/web-application.yml
```diff
@@ -0,0 +1,257 @@
...
+          Type: forward
+      Conditions:
+        - Field: host-header
+          Values:
+            - !Sub ${AppName}.*
+      ListenerArn: !Ref PlatformALBHTTPListenerArn
+      Priority: !Join ["", [!Ref PlatformALBListenerPriorityPrefix, 03]]
+  ALBTargetGroup500Alarm:
+    Type: "AWS::CloudWatch::Alarm"
+    # Condition: CreateProductionResources
```
farski: Are you creating these in staging for any particular reason? If you do want staging alarms I don't think they should go to the `Error` topic; probably `Debug` or `Info`

stacks/web-application.yml
```diff
@@ -0,0 +1,257 @@
...
+  WebService:
+    Type: "AWS::ECS::Service"
+    Properties:
+      Cluster: !Ref ECSCluster
+      DeploymentConfiguration:
+        MaximumPercent: 200
+        MinimumHealthyPercent: 50
+      DesiredCount: 2
+      LoadBalancers:
+        - ContainerName: !Sub ${AppName}-container
```
farski: Actually on L180 this is called `-web`, maybe this is a typo?

stacks/root.yml
```diff
@@ -332,6 +334,49 @@ Resources:
         PlaySecretsVersion: !Ref PlaySecretsVersion
       TemplateURL: !Join ["", ["http://s3", !If [IsUsEast1, "", "-"], !If [IsUsEast1, "", !Ref "AWS::Region"], ".amazonaws.com/", "Fn::ImportValue": !Sub "${InfrastructureStorageStackName}-InfrastructureSourceBucket", "/", !Ref InfrastructureGitCommit, "/stacks/play.prx.org.yml"]]
       TimeoutInMinutes: "5"
+  ##############################################################################
```
farski: I've only done a cursory review of these changes so far; focused on the other template for this review

stacks/web-application.yml
```diff
@@ -0,0 +1,257 @@
...
+  # Shared ENV #################################################################
+  # App ECS #################################################################
+  EcrImageTag:
+    Type: String
+  # App ENV #################################################################
+  AppName:
+    Type: String # castle, crier, cms, feeder
+  SecretsVersion:
+    Type: String
+  CreateWorker:
```
kookster: I followed example of other stacks that use "# Castle ENV #..." or "# Publich ENV #...", but I made it generic - suggested change?

stacks/web-application.yml
```diff
@@ -0,0 +1,257 @@
...
+    Type:  String
+  ContainerPort:
+    Type: String
+    Default: 3000
+  ContainerMemory:
```
kookster: currently yes, the examples I looked at have these as the same already. Generally the amount of memory and CPU will be similar since it is running the same app and docker image. We can break these out if needed later.

stacks/web-application.yml
kookster: good point, I'll add a condition for only production.

stacks/web-application.yml
kookster: will fix, thanks

stacks/web-application.yml
```diff
@@ -0,0 +1,257 @@
...
+  # Misc #######################################################################
+  OpsDebugMessagesSnsTopicArn:
+    Type: String
+  OpsErrorMessagesSnsTopicArn:
+    Type: String
+  EnvironmentType:
+    Type: String
+  EnvironmentTypeAbbreviation:
+    Type: String
+  EcrRegion:
```
kookster: done.

stacks/web-application.yml
farski: That parameter only gets used in the template afaik, so maybe just move it up in to _Misc_

stacks/web-application.yml
farski: Sounds good

stacks/web-application.yml
kookster: I think 500s should general be an error, perhaps we can leave this for now and see if it is a thing we need to tune.

stacks/root.yml
kookster: ok, made minor change to this as well, only to put `Ecr` params together.

kookster:
I think I made the necessary changes, I'm going to try testing this stack

PR #75, opened 2017-06-27
Include branch name in CI

cavis:
Also pass the branch name to CI codebuild, so our [codecov fork](https://github.com/PRX/codecov-node) can work without the `.git` directory.

### DIFF
--- a/ci/lambdas/github-event-handler/index.js
+++ b/ci/lambdas/github-event-handler/index.js
@@ -291,8 +291,12 @@ function triggerBuild(versionId, event, callback) {
 
     if (event.pull_request) {
         const num = event.pull_request.number;
+        const branch = event.pull_request.head.ref;
         environmentVariables.push({ name: 'PRX_GITHUB_PR', value: `${num}` });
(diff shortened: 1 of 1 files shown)


ci/lambdas/github-event-handler/index.js
```diff
@@ -291,8 +291,12 @@ function triggerBuild(versionId, event, callback) {
 
     if (event.pull_request) {
         const num = event.pull_request.number;
+        const branch = event.pull_request.head.ref;
         environmentVariables.push({ name: 'PRX_GITHUB_PR', value: `${num}` });
+        environmentVariables.push({ name: 'PRX_BRANCH', value: branch});
     } else {
+        const branch = (event.ref || 'unknown').replace(/^refs\/heads\//, '');
```
farski: Are there known situations where there's no `ref`?

cavis: I haven't seen any - was mostly just covering all the bases here.  If we happen to add some other notification types down the road.

PR #77, opened 2017-06-30
add feeder-web to Infrastructure

gcampo88:
Adding feeder-web to Infrastructure. 

I believe that the feeder.prx.org repo is already set up for migration, so the next steps will be:
- [ ] getting feeder secrets into the staging secrets bucket
- [ ] updating the template staging config with the secrets version + ecr image tag 

(More info on steps [here](https://github.com/PRX/prx.org/wiki/Migrating-backend-apps-to---PRX-CICD))

### DIFF
--- a/stacks/root.yml
+++ b/stacks/root.yml
@@ -432,7 +432,7 @@ Resources:
         EcrImageTag: !Ref FeederEcrImageTag
         SecretsVersion: !Ref FeederSecretsVersion
         AppName: "feeder"
-        CreateWorker: "true"
+        CreateWorker: "false"
         ContainerPort: 3000
         ContainerMemory: 500
         ContainerCpu: 128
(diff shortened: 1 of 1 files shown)


stacks/root.yml
```diff
@@ -86,6 +90,8 @@ Mappings:
       prefix: 20
     Castle:
       prefix: 30
+    Feeder:
+      prefix: 32
```
farski: Feeder API I think gets hit a lot, so probably it should be very high priority

cavis: I would actually put Feeder as the top priority.  It gets a ton of traffic from Dovetail.  And apparently it costs us real life money for every rule a request is checked against.

gcampo88: 👍

PR #82, opened 2017-07-12
Shared ECS + ALB stack for feeder

cavis:
Give feeder an ALB.  Mostly copied out of dovetail and the other shared stack.

### DIFF
--- a/stacks/web-alb-application.yml
+++ b/stacks/web-alb-application.yml
@@ -138,12 +138,12 @@ Resources:
           Value: !Sub ${AppName}-${EnvironmentType}
       VpcId: !Ref VPC
   # CloudWatch Alarms
-  ALB500Alarm:
+  ALBTarget500Alarm:
     Type: "AWS::CloudWatch::Alarm"
     Condition: CreateProductionResources
     Properties:
       ActionsEnabled: true
-      AlarmName: !Sub ${AppName} ALB HTTP 5XXs
+      AlarmName: !Sub ${AppName} Server 5XXs
       AlarmActions:
         - !Ref OpsErrorMessagesSnsTopicArn
       InsufficientDataActions:
@@ -151,7 +151,7 @@ Resources:
       OKActions:
         - !Ref OpsErrorMessagesSnsTopicArn
       AlarmDescription: !Sub |
-        5XX server errors originating from the ${AppName} load balancer exceeded 0
+        5XX server errors originating from the ${AppName} application
       ComparisonOperator: GreaterThanThreshold
       EvaluationPeriods: "1"
       MetricName: HTTPCode_Target_5XX_Count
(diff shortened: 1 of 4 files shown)


stacks/web-alb-application.yml
```diff
@@ -0,0 +1,276 @@
...
+      ActionsEnabled: true
+      AlarmName: !Sub ${AppName} ALB HTTP 5XXs
+      AlarmActions:
+        - !Ref OpsErrorMessagesSnsTopicArn
+      InsufficientDataActions:
+        - !Ref OpsErrorMessagesSnsTopicArn
+      OKActions:
+        - !Ref OpsErrorMessagesSnsTopicArn
+      AlarmDescription: !Sub |
+        5XX server errors originating from the ${AppName} load balancer exceeded 0
```
gcampo88: might vote that we update the alarm description for the platform alarm to specify that it's the platform load balancer to be extra clear --
```YAML
# load-balancers.yml
AlarmDescription: >
        5XX server errors originating from the *platform* ALB exceeded 0
```

stacks/web-alb-application.yml
```diff
@@ -0,0 +1,276 @@
...
+      ComparisonOperator: GreaterThanThreshold
+      EvaluationPeriods: "1"
+      MetricName: HTTPCode_Target_5XX_Count
```
gcampo88: do you want target or elb? (the platform ALB uses `HTTPCode_ELB_5XX_Count`). by the description, i'm assuming you want ELB. 

docs say that target will give you: `The number of HTTP response codes generated by the targets. This does not include any response codes generated by the load balancer.` versus elb that gives you: `The number of HTTP 5XX server error codes that originate from the load balancer. This count does not include any response codes generated by the targets.`

stacks/web-alb-application.yml
```diff
@@ -0,0 +1,276 @@
...
+        - CertificateArn: !Ref VPCCertificateArn
+      LoadBalancerArn: !Ref ALB
+      Port: 443
+      Protocol: HTTPS
+      DefaultActions:
+        - Type: forward
+          TargetGroupArn: !Ref ALBTargetGroup
+  ALBTargetGroup:
+    Type: "AWS::ElasticLoadBalancingV2::TargetGroup"
+    DependsOn: ALB
```
cavis: The shared ALB probably isn't a good example, since it's in a completely separate stack from those apps.  Which I think requires the entire load-balancers stack to complete before it even starts creating those app-specific target groups.

I may have gotten a bit over-zealous with `DependsOn` in the Dovetail stack, which this is copied from.  Was having some order-of-creation issues there.  Let me re-visit some of these.

stacks/web-alb-application.yml
cavis: +1

stacks/web-alb-application.yml
cavis: In this case, target.  That platform 'ELB_5XX` alarm was to catch any requests that fall all the way through host-based routing.  (Which we really shouldn't get anymore, since fourohfour catches them).

That said, we do get a fair number of 503/504s from the dovetail load balancer when tasks are spinning up/down a lot, so it's certainly worth monitoring.  Will add something, and take a guess at what the threshold should be.

cavis:
@gcampo88 back at ya.

PR #86, opened 2017-07-14
Add a mostly-random nonce to Root stack for utility purposes

farski:
For custom resources it can be useful to have a value that changes each
time the pipeline runs. This uses the S3 object key for an artifact
that is always created, and therefore should be unique to each pipeline
execution


farski:
CD stack need to be updated manually for this change. Root deploys will complain about params until that happens.

Issue #68, opened 2017-06-07
Migrate ID Staging


cavis:
ID Production currently lives on contegix.  But we can migrate ID-staging!

- [x] Setup aws-secrets
  - [x] Modify `bin/application` to _optionally_ read aws-secrets
  - [x] Setup staging secrets in S3 bucket
- [x] Add `.prxci`
- [x] Add prx app label to Dockerfile
- [x] Add to Infrastructure CD
- [ ] Bonus points: meta.prx.org acceptance tests

pkarman:
What is the status of this ticket? @cavis @kookster

cavis:
It's done!  http://id.staging.prx.tech/api/v1/certs

PR #91, opened 2017-07-24
Add ID service

cavis:
See #68.

PRX/id.prx.org#36 has to get merged first.

### DIFF
--- a/stacks/root.yml
+++ b/stacks/root.yml
@@ -55,6 +55,10 @@ Parameters:
     Type: String
   FeederSecretsVersion:
     Type: String
+  IdEcrImageTag:
+    Type: String
+  IdSecretsVersion:
+    Type: String
   PlayEcrImageTag:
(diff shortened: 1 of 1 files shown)


stacks/root.yml
```diff
@@ -103,6 +107,8 @@ Mappings:
       prefix: 15
     Play:
       prefix: 21
+    Id:
```
gcampo88: this order seems reasonable, but out of curiosity, how often does ID get hit? publish and prx.org go through it, right? what else?

cavis: Any place that you're logged in, users will have to hit ID.  So users of Publish/Castle/Adze/Metrics and eventually V3.  My thinking is that CMS and Play get hit by unauthorized requests, so they go first.  Play will likely get way more traffic than CMS, but most of that is CloudFront, so I left it 2nd here.

PR #98, opened 2017-08-08
Drain ECS tasks before terminating

### LINKED ISSUE #96: Automate ECS instance draining
Currently when an ASG scales down, ECS tasks running on those instances are just terminated, leaving you with several ALB 500s until the health check failures remove that instance from the target group.

Use lifecycle hooks + lambda to safely drain the instance before it terminates -> https://github.com/awslabs/ecs-cid-sample

cavis:
Fixes #96.

Use a lifecycle hook to drain ECS tasks before an instance terminates.  Prevents ALB 500s while it's waiting for the health-check to fail.

See https://github.com/awslabs/ecs-cid-sample

### DIFF
--- a/stacks/ecs.yml
+++ b/stacks/ecs.yml
@@ -446,6 +446,12 @@ Resources:
           NotificationTypes:
             - "autoscaling:EC2_INSTANCE_LAUNCH_ERROR"
             - "autoscaling:EC2_INSTANCE_TERMINATE_ERROR"
+        - TopicARN: !Ref ECSDrainSNSTopic
+          NotificationTypes:
+            - "autoscaling:EC2_INSTANCE_LAUNCH"
+            - "autoscaling:EC2_INSTANCE_LAUNCH_ERROR"
+            - "autoscaling:EC2_INSTANCE_TERMINATE"
+            - "autoscaling:EC2_INSTANCE_TERMINATE_ERROR"
       Tags:
         - Key: Project
           Value: platform.prx.org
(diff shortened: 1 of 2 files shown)


stacks/ecs.yml
```diff
@@ -462,6 +468,104 @@ Resources:
...
+            Action:
+              - "sts:AssumeRole"
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/service-role/AutoScalingNotificationAccessRole
+      Path: "/"
+  ECSDrainLambdaRole:
+    Type: "AWS::IAM::Role"
+    Properties:
+      Policies:
+        -
```
farski: We haven't been formatting YAML lists with this extra line

stacks/ecs.yml
```diff
@@ -462,6 +468,104 @@ Resources:
...
+          PolicyName: "drain-lambda-inline"
+          PolicyDocument:
+            Version: "2012-10-17"
+            Statement:
```
farski: I wouldn't hate if this gotten broken up a bit with more granular resource allowances, at least for the actions that aren't GET-like (:submit*, :update*, and :publish*, I guess)

cavis:
Updated.  I think.

stacks/ecs.yml
```diff
@@ -462,6 +468,112 @@ Resources:
...
+                  - ecs:DescribeTasks
+                  - ecs:ListContainerInstances
+                  - ecs:ListTasks
+                  - sns:ListSubscriptions
+              - Effect: "Allow"
+                Resource: !Ref ECSClusterASG
+                Action:
+                  - autoscaling:CompleteLifecycleAction
+              - Effect: "Allow"
+                Resource: !Ref ECSCluster
```
farski: May be the case that some of these , like the task state change, need access to subresources of the ECSCluster ARN. Couldn't find anythign obvious in the docs, so…we'll find out?

PR #102, opened 2017-08-28
DNS record templates

farski:
Adds templates for Route 53 hosted zones for several domains that already exist in Route 53, as well as prx.org, which currently is hosted on Contegix's name servers.


pkarman:
@farski who should review this?

farski:
It's mostly immaterial, as I've already deployed the changes that are included in this PR. Mostly just holding off on the merge until we finish the review of which records should be included, and the comparison of current zone to the new zone.

PR #113, opened 2017-09-07
Disable audiogram stack

farski:


farski:
Will remove the code and clean up secrets/ECR in other tickets

Issue #110, opened 2017-09-05
Dovetail multiple-tasks-per-instance


cavis:
Allow DT to run multiple tasks per instance (currently limited to 1-per-instance).  This simplifies all sorts of things (auto-scaling, ECS-instance counts, etc).

I _think_ this will involve recreating the service, which may or may not result in an outage while the ALB moves the route from one target group to another.  Will need to very carefully test this out in staging.

farski:
@cavis Remind me what's keeping this from happening right now?

cavis:
Originally, we were seeing network issues trying to run > 1 dovetail on a `t2` instance.  I'm _guessing_ that some of that was us burning through CPU credits too fast under load.  Now that we're on non-burst instances, I'm hoping that won't be a problem.  (Though we should be careful about testing that).

The only thing stopping us from testing is the possibility that CFN will tear down too much of dovetail in order to create a new service without the "1 per instance" placement rule.

(Probably a few other misc updates to make around the DT-monitoring/killing lambdas, and making sure we have enough disk space for that many tasks).

cavis:
Deployed, updated lambdas/dashboards, and seems to be working fine.  Calling this good.

PR #119, opened 2017-10-02
Unconstrained dovetail

cavis:
Working towards #110.  Spin up another dovetail service/target-group, without the 1-per-instance constraint on it.

After this finishes, will change the target-group under the http/https listeners.  And watch for errors for awhile before tearing down the old service/target.

(This should enable a very quick rollback, if need be).


farski:
AKA Dovetail off the chain

PR #122, opened 2017-10-03
adds metrics

sandikbarr:
adds metrics?

### DIFF
--- a/stacks/root.yml
+++ b/stacks/root.yml
@@ -116,12 +116,12 @@ Mappings:
       prefix: 28
     Castle:
       prefix: 35
+    Metrics:
+      prefix: 36
     Jingle:
       prefix: 38
     Crier:
       prefix: 42
-    Metrics:
-      prefix: 36
 Resources:
   VPCStack:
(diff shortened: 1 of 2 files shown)


stacks/root.yml
```diff
@@ -116,6 +120,8 @@ Mappings:
       prefix: 38
     Crier:
       prefix: 42
+    Metrics:
```
gcampo88: Metrics should have a unique # here (jingle has already taken 38). Also good to maintain order in increasing listener priority. My guess is Metrics should be somewhere around -- 36? Lower priority than castle (which will get hit at least as often as metrics) but higher than jingle (which will only be the shows we sell ads for)

sandikbarr: done!

gcampo88: last nitpicky thing -- good to maintain this list in order of the priority prefix, so just move this to below castle / above jingle and then i think all set!

gcampo88:
LGTM! Will hold off on merging til we're ready to deal with the consequences :). 🥇 💯 📊

Issue #10, opened 2017-02-17
ECS ASG scaling rules


farski:
After #120, add some scaling rules to the prod ASG based on CPU/Memory reservation.  Something like ... always have 1 at least half-empty instance around.

Will also require some CFN changes to prevent resetting the ASG DesiredCount every time we deploy.

Then, can tackle #134 and #135.

kookster:
Estimating getting this figured out for fancy prod cluster, for dovetail and feeder apps.

PR #141, opened 2017-10-16
ASG sizing lambda

cavis:
See #10.  Attempts to scale an auto-scaling-group to most effectively fit an ECS cluster.

The 3 "threshold" ENVs (cpu/memory/count) configure a theoretical task that you always want to leave room for on the cluster.  It should have reservations at least as big as the largest task running on your cluster.

For instance, if you set those to 700/1400/2, this lambda would make sure there was always room for 2 more tasks on the cluster, with 700 cpu and 1400 memory.  If not, it scales up by 1.

Scaling down is a bit more complex.  You can't know which instance/tasks the ASG will kill when scaling down, so just assume it's completely full of these big "threshold" tasks.  So if 1 instance can hold 4 threshold tasks ... and I need space for 2 more tasks so we don't immediately scale back up ... there must be at least 6 slots available in the cluster in order for us to scale down.

### DIFF
--- a/notifications/lambdas/slack-message-relay/index.js
+++ b/notifications/lambdas/slack-message-relay/index.js
@@ -17,7 +17,11 @@ exports.handler = (event, context, callback) => {
         const sns = event.Records[0].Sns;
 
         const attrs = sns.MessageAttributes;
-        const webhookUrl = attrs.WebhookURL.Value;
+
+        let webhookUrl = process.env['DEFAULT_WEBHOOK_URL'];
+        if (attrs.WebhookURL) {
+          webhookUrl = attrs.WebhookURL.Value;
+        }
 
         // Setup request options
         const options = url.parse(webhookUrl);
--- a/notifications/lambdas/slack-message-relay/index.js
+++ b/notifications/lambdas/slack-message-relay/index.js
@@ -18,7 +18,7 @@ exports.handler = (event, context, callback) => {
 
         const attrs = sns.MessageAttributes;
 
-        let webhookUrl = process.env['DEFAULT_WEBHOOK_URL'];
+        let webhookUrl = process.env.DEFAULT_WEBHOOK_URL;
         if (attrs.WebhookURL) {
           webhookUrl = attrs.WebhookURL.Value;
         }
--- a/stacks/ecs.yml
+++ b/stacks/ecs.yml
@@ -588,11 +588,11 @@ Resources:
         - Key: Project
           Value: Infrastructure
       Timeout: 300
-  ECSScaleRole:
+  ASGSizerRole:
     Type: "AWS::IAM::Role"
     Properties:
       Policies:
-        - PolicyName: "scale-lambda-inline"
+        - PolicyName: "ASGSizerLambdaPolicy"
           PolicyDocument:
             Version: "2012-10-17"
             Statement:
@@ -603,23 +603,19 @@ Resources:
                   - logs:CreateLogStream
                   - logs:PutLogEvents
                   - ecs:DescribeContainerInstances
+                  - ecs:DescribeServices
+                  - ecs:DescribeTaskDefinition
                   - ecs:ListContainerInstances
+                  - ecs:ListServices
               - Effect: "Allow"
                 Resource: "*" # TODO: how to get an ASG ARN?
                 Action:
                   - autoscaling:DescribeAutoScalingGroups
                   - autoscaling:SetDesiredCapacity
-      AssumeRolePolicyDocument:
-        Version: "2012-10-17"
-        Statement:
-          - Effect: "Allow"
-            Principal:
-              Service:
-                - "lambda.amazonaws.com"
-            Action:
-              - "sts:AssumeRole"
+      ManagedPolicyArns:
+        - "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
       Path: "/"
-  ECSScaleFunction:
+  ASGSizerFunction:
     Type: "AWS::Lambda::Function"
     Properties:
       Code:
@@ -627,14 +623,14 @@ Resources:
           Fn::ImportValue:
             !Sub ${InfrastructureStorageStackName}-InfrastructureSupportBucket
         S3Key: utility/ecs-asg-sizer.zip
-      Description:
+      Description: Make an ASG the perfect size for an ECS cluster
       Environment:
         Variables:
           ASG_NAME: !Ref ECSClusterASG
           ECS_CLUSTER: !Ref ECSCluster
       Handler: lambda_function.lambda_handler
       MemorySize: 128
-      Role: !GetAtt ECSScaleRole.Arn
+      Role: !GetAtt ASGSizerRole.Arn
       Runtime: python3.6
       Tags:
         - Key: Project
--- a/stacks/ecs.yml
+++ b/stacks/ecs.yml
@@ -49,12 +49,6 @@ Parameters:
     Type: String
   SecretsInstanceAccessPolicyArn:
     Type: String
-  ThresholdCpu:
-    Type: String
-  ThresholdMemory:
-    Type: String
-  ThresholdCount:
-    Type: String
 Mappings:
   EnvironmentTypeMap:
     Testing:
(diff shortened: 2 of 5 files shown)


stacks/ecs.yml
```diff
@@ -588,6 +594,61 @@ Resources:
...
+                Action:
+                  - logs:CreateLogGroup
+                  - logs:CreateLogStream
+                  - logs:PutLogEvents
+                  - ecs:DescribeContainerInstances
+                  - ecs:ListContainerInstances
+              - Effect: "Allow"
+                Resource: "*" # TODO: how to get an ASG ARN?
+                Action:
+                  - autoscaling:DescribeAutoScalingGroups
```
cavis: Not actually sure this is right.  Guess we'll find out!

stacks/root.yml
```diff
@@ -98,6 +98,15 @@ Mappings:
       abbreviation: stag
     Production:
       abbreviation: prod
+  ASGScalingMap:
```
cavis: Made sense to include these numbers "near" the ECSReservationMap, since they're related and should be kept in-sync.

cavis:
My output in staging looks something like:

```
[DEBUG] Targeting threshold 2x 400 cpu / 1000 mem
[DEBUG]     i-09d534950463cdbb0 - 448 / 201
[DEBUG]     i-02961d7d4fa38bd7d - 896 / 1601
[DEBUG]     i-0603ea1ca85e0cb11 - 1024 / 2001
[DEBUG]     i-0759562af5853ee07 - 384 / 1
[DEBUG]     i-0258b2b66eaba27fe - 448 / 201
[DEBUG]     i-080bd5ea6ccd55072 - 896 / 1601
[DEBUG]     i-08a6ac98730a79ac6 - 896 / 1601
[DEBUG]     i-0c111031236098a8d - 832 / 1401
[DEBUG]     i-07ca2ee58264b12c5 - 384 / 1
[INFO] Scale DOWN: 6 slots available (3/instance)
[DEBUG] Updated desired capacity to 8
```

farski:
What do you think about changing the termination policy to `NewestInstance`? We'd lose the default property that the group tries to spread across AZ's, but it seems likely that once we start autoscaling more regularly, the newest instances will be the ones with the tasks necessary for the spike and the ones we want to lose first (or care least about losing). Would likely/hopefully avoid the situation where container instances tear down needed tasks unrelated to scaling.

cavis:
Good point.  We should probably try that out and see what happens.  I was also thinking their might be some chaos-monkey-esque advantages to using the "oldest" policy.  Which would make sure no EC2 instances were around for very long.  But as you said, if we're scaling a lot, that could make things more unstable.  (Dovetail, for instance, only performs really well once its local file-cache is primed).

stacks/root.yml
farski: Does it make any sense to abstract the reservations into like small/medium/large and just literally use the same value references of large for the threshold?

utility/lambdas/ecs-asg-sizer/lambda_function.py
```diff
@@ -0,0 +1,115 @@
+import boto3
```
farski: Have not reviewed the lambda code yet

farski: Make sure this is following the newish linting style; details in Contributing.md

utility/lambdas/ecs-asg-sizer/lambda_function.py
```diff
@@ -0,0 +1,115 @@
...
+import json
+import os
+
+session = boto3.session.Session()
```
farski: I think we've mostly been doing under_score naming for all functions, variables, etc since it's all new code and that's what the convention is for python

utility/lambdas/ecs-asg-sizer/lambda_function.py
```diff
@@ -0,0 +1,115 @@
...
+    asgName = get_env('ASG_NAME')
+    clusterName = get_env('ECS_CLUSTER')
+    thresholdCpu = int(get_env('THRESHOLD_CPU'))
+    thresholdMemory = int(get_env('THRESHOLD_MEMORY'))
+    thresholdCount = int(get_env('THRESHOLD_COUNT'))
+
+    usages = get_usages(clusterName)
+    log_debug('Targeting threshold {0}x {1} cpu / {2} mem'.format(thresholdCount, thresholdCpu, thresholdMemory))
+    for usage in usages:
+        log_debug('    {0} - {1} / {2}'.format(usage['id'], usage['remainingCpu'], usage['remainingMemory']))
```
farski: We can do interpolation in py3.6 functions which make these less gross
https://cito.github.io/blog/f-strings/

utility/lambdas/ecs-asg-sizer/lambda_function.py
```diff
@@ -0,0 +1,115 @@
...
+    slotsPerInstance = 0
+    while totalCpu > 0 and totalMemory > 0:
+        slotsPerInstance += 1
+        totalCpu -= thresholdCpu
+        totalMemory -= thresholdMemory
+
+    # scale!
+    msg = '{0} slots available ({1}/instance)'.format(slotsAvailable, slotsPerInstance)
+    if slotsAvailable < thresholdCount:
+        log_info('Scale UP: ' + msg)
```
farski: We should standardize around up/down vs in/out. My preference would be in/out as that's what AWS seems to be using recently

utility/lambdas/ecs-asg-sizer/lambda_function.py
```diff
@@ -0,0 +1,115 @@
...
+def scale_asg(asgName, delta):
+    group = asgClient.describe_auto_scaling_groups(AutoScalingGroupNames=[asgName])['AutoScalingGroups'][0]
+    desired = group['DesiredCapacity'] + delta
+    if desired > group['MaxSize']:
+        log_warn('Already at max-size')
+    elif desired < group['MinSize']:
+        log_warn('Already at min-size')
+    else:
+        asgClient.set_desired_capacity(AutoScalingGroupName=asgName, DesiredCapacity=desired, HonorCooldown=False)
+        log_debug('Updated desired capacity to {0}'.format(desired))
```
farski: The logic here I'll need to spend more time with later today

stacks/root.yml
cavis: I think I've actually figured out a better way to do this.  Instead of ENV-ing it, just have the lambda ask aws for a list of all service-task-defs in the cluster, and pull out the largest CPU/Memory reservations there.  Less configs to keep in-sync.

utility/lambdas/ecs-asg-sizer/lambda_function.py
cavis: Good call.  I haven't actually ever read anything about python styling!

utility/lambdas/ecs-asg-sizer/lambda_function.py
cavis: Just figured out i'm running 2.7 locally.  Whoops!

stacks/root.yml
farski: Ooo I like that

cavis:
Okay @farski - give that another look.  Let's call this PR good, and I'll start working on another one to figure out how/what slack notifications I want.  (For now, logging to CW will be good enough to start testing this).

If this actually works, I'll also do another ecs.yml PR to put it on a cron.

notifications/lambdas/slack-message-relay/index.js
```diff
@@ -17,7 +17,11 @@ exports.handler = (event, context, callback) => {
         const sns = event.Records[0].Sns;
 
         const attrs = sns.MessageAttributes;
-        const webhookUrl = attrs.WebhookURL.Value;
+
+        let webhookUrl = process.env['DEFAULT_WEBHOOK_URL'];
```
farski: Linting will probably complain about not using dot notation

stacks/ecs.yml
```diff
@@ -588,6 +590,64 @@ Resources:
...
+    Properties:
+      Policies:
+        - PolicyName: "ASGSizerLambdaPolicy"
+          PolicyDocument:
+            Version: "2012-10-17"
+            Statement:
+              - Effect: "Allow"
+                Resource: "*"
+                Action:
+                  - logs:CreateLogGroup
```
farski: These log actions are dupes with the `AWSLambdaBasicExecutionRole` being included

PR #153, opened 2017-11-06
Codebuild secrets

cavis:
Give codebuild access to aws-secrets.  (Assumes names of managed policies will stay the same).

### DIFF
--- a/ci/ci.yml
+++ b/ci/ci.yml
@@ -154,6 +154,9 @@ Resources:
                   - "ecr:PutImage"
                 Resource:
                   - !Sub "arn:aws:ecr:us-east-1:${AWS::AccountId}:repository/*"
+      ManagedPolicyArns:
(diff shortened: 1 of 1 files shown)


ci/ci.yml
```diff
@@ -154,6 +154,9 @@ Resources:
                   - "ecr:PutImage"
                 Resource:
                   - !Sub "arn:aws:ecr:us-east-1:${AWS::AccountId}:repository/*"
+      ManagedPolicyArns:
+        - !Sub "arn:aws:iam::${AWS::AccountId}:policy/key-prx-infrastructure-${AWS::Region}-secrets-instance-access-policy"
+        - !Sub "arn:aws:iam::${AWS::AccountId}:policy/key-prx-infrastructure-${AWS::Region}-secrets-instance-decrypt-policy"
```
farski: These are exported from the Secrets stack, so I would think they could just be imported here. You'd just need to add a parameter to the CI template for the name of the secrets stack, like how storage and notifications are already.

cavis:
@farski thusly?

PR #163, opened 2017-12-01
Analytics ingest lambdas

cavis:
Creates the 3 analytics lambdas, for bigquery/pingbacks/redis.

- Not sure the alarms are what we actually want
- Does not include any log metrics (tend to use these more than error counts in our monitoring)
- Does not include any actual secret ENVs
  - Will try adding these manually, and check if CFN blows them away on subsequent deploys
- Does not include any triggers
  - Don't actually want kinesis triggers in CFN, yet

### DIFF
--- a/stacks/analytics-ingest-lambda.yml
+++ b/stacks/analytics-ingest-lambda.yml
@@ -119,7 +119,6 @@ Resources:
       Dimensions:
         - Name: FunctionName
           Value: !Ref AnalyticsPingbacksLambdaFunction
-          Value: !Ref AnalyticsBigqueryLambdaFunction
   AnalyticsRedisErrorAlarm:
     Type: "AWS::CloudWatch::Alarm"
     Properties:
--- a/stacks/analytics-ingest-lambda.yml
+++ b/stacks/analytics-ingest-lambda.yml
@@ -1,0 +1,142 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Description: Analytics ingest lambda functions
+Parameters:
+  OpsErrorMessagesSnsTopicArn:
+    Type: String
+  CodeS3Bucket:
+    Type: String
+  CodeS3ObjectVersion:
+    Type: String
+  EnvironmentTypeAbbreviation:
+    Type: String
+Resources:
+  # IAM Roles
+  AnalyticsLambdaExecutionIAMRole:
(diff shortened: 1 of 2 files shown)


stacks/analytics-ingest-lambda.yml
```diff
@@ -0,0 +1,142 @@
...
+      Code:
+        S3Bucket: !Ref CodeS3Bucket
+        S3Key: !Sub "lambda/PRX-analytics-ingest-lambda.zip"
+        S3ObjectVersion: !Ref CodeS3ObjectVersion
+      Description: Dovetail analytics to BigQuery
+      Environment:
+        Variables:
+          BQ_IMPRESSIONS_TABLE: "impressions"
+          BQ_DOWNLOADS_TABLE: "downloads"
+          BQ_PROJECT_ID: "prx-metrics"
```
kookster: Do we vary any of this by env? - Is this the same project for prod and staging?

stacks/analytics-ingest-lambda.yml
```diff
@@ -0,0 +1,142 @@
...
+      MetricName: Errors
+      Namespace: AWS/Lambda
+      Period: "300"
+      Statistic: Sum
+      Threshold: "0"
+      TreatMissingData: notBreaching
+      Dimensions:
+        - Name: FunctionName
+          Value: !Ref AnalyticsPingbacksLambdaFunction
+          Value: !Ref AnalyticsBigqueryLambdaFunction
```
kookster: you need both these in here? for the pingback error alarm?

stacks/analytics-ingest-lambda.yml
cavis: We use the same "project" for all 3 of our datasets: development, staging and production.

It's actually a bit of a TBD if this will work - having _some_ of the ENVs in cloudformation, but excluding others.  My hope is that CFN will just leave those ENVs alone that it doesn't know about.  We'll find out after this deploys.

The long-term hope is that Lambda gives 1st-class-citizen status to parameter-store.  Then we can just specify a paramstore name in here, and vary it based on staging-vs-prod.

stacks/analytics-ingest-lambda.yml
cavis: Whoops!  Typo.

PR #164, opened 2017-12-06
Don't publish pull requests

farski:


ci/utility/post_build.sh
```diff
@@ -148,8 +141,8 @@ init() {
         build_error "A previous CodeBuild phase did not succeed"
     fi
 
-    # Handle code publish if enabled
-    if [ -n "$PRX_CI_PUBLISH" ]
+    # Handle code publish if enabled and not pull requests
+    if [ -n "$PRX_CI_PUBLISH" ] && [-z "$PRX_GITHUB_PR"]
```
farski: @cavis Please confirm that this is a thing. I also saw people using `-a` and couldn't figure out the difference

PR #167, opened 2017-12-07
Include PRX_BRANCH in ci

cavis:
Anything using codecov will also need to know which branch is being built.


farski:
Don't forget to publish the code change to the Lambda if you haven't already

PR #176, opened 2017-12-28
Add response time alarm to ALB app template

farski:


stacks/web-alb-application.yml
```diff
@@ -217,6 +219,32 @@ Resources:
...
+      AlarmDescription: !Sub |
+        Target response time value for ${AppName} load balancer targets at 95th
+        percentile higher than expected
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      EvaluationPeriods: "1"
+      ExtendedStatistic: p95
+      MetricName: TargetResponseTime
+      Namespace: AWS/ApplicationELB
+      Period: "300"
+      Threshold: "0.1"
```
gcampo88: Making sure I understand the role of `threshold` here -- so we're declaring that the expected response time for this ALB is 0.1 seconds, and if any response in a 5 min period is greater than or equal to 95% above that expected response, the alarm will activate? Or is the expected response time determined elsewhere? (I'm a little confused looking at the similar alarms in Dovetail, since it looks like the p99 goes off way more often than the p95, which I would not have expected --? )

PR #181, opened 2018-02-12
Explicit ECS AMI names

cavis:
Allows upgrading staging separately from prod.

### DIFF
--- a/stacks/ecs.yml
+++ b/stacks/ecs.yml
@@ -60,6 +60,7 @@ Mappings:
     Testing:
       InstanceType: t2.micro
       InstanceStorage: 22
+      AmiName: "amzn-ami-2017.03.d-amazon-ecs-optimized"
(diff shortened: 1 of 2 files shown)


stacks/ecs.yml
```diff
@@ -141,7 +144,7 @@ Resources:
     Properties:
       ServiceToken: !GetAtt ECSAmiLookupFunction.Arn
       Region: !Ref AWS::Region
-      Rotation: 1 # Change this to trigger AMI updates
+      AmiName: !FindInMap [EnvironmentTypeMap, !Ref EnvironmentType, AmiName]
```
cavis: I'm _hoping_ that this will do the same thing as rotation ... only call the Lambda when the `AmiName` value changes.

farski: My understanding is that it should work that way

PR #183, opened 2018-02-26
Dovetail Cloudfront

cavis:
A distribution to pull stitched files from the Dovetail ALB.

Will require manually adding the Alias record to `prxu.org-hosted_zone.yml` after deploying.


cavis:
Moving on to the Lambda-ized version, in another PR.

PR #189, opened 2018-03-01
Dovetail stitch lambda

cavis:
Requires merge of PRX/dovetail-stitch-lambda#2


farski:
This had better have all the proper tagging…

PR #191, opened 2018-03-01
CORS Allow-Headers should be wildcard

pkarman:
See https://github.com/PRX/publish.prx.org/issues/484#issuecomment-369730170


farski:
My understanding is that wildcards for `Access-Control-Allow-Headers` aren't supported

pkarman:
supported in the spec, but not yet in browsers. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers#Compatibility_notes

So I will change this to whitelist the offending header.

PR #192, opened 2018-03-01
Add 5xx error alarm to DT CDN

farski:
@cavis feel free to mess with the numbers


farski:
ack I forgot to pass in the sns topics

PR #193, opened 2018-03-02
Feat/staging deltas

farski:


cd/cd.yaml
```diff
@@ -528,14 +528,14 @@ Resources:
...
-            - Name: StagingDeploy
+            - Name: CreateStagingChangeSet
               ActionTypeId:
                 Category: Deploy
                 Owner: AWS
                 Provider: CloudFormation
                 Version: 1
               Configuration:
-                ActionMode: REPLACE_ON_FAILURE
+                ActionMode: CHANGE_SET_REPLACE
```
cavis: Is this an unrelated change?

cavis: Or is this just to separate creating/executing the changeset, so you can notify in between?

farski: oh hm

farski: that probably should change back. replace is just a more aggressive method, so if staging were to get stuck in a bad way another deploy could just replace it, rather than trying to update which always needs to happen from a stable state

PR #194, opened 2018-03-02
Add CORS headers to GET response

pkarman:
I added the minimum amount I thought I could get away with. Not sure how to test this at all.

### DIFF
--- a/stacks/upload.prx.org.yml
+++ b/stacks/upload.prx.org.yml
@@ -42,6 +42,18 @@ Resources:
         Type: AWS_PROXY
         IntegrationHttpMethod: POST
         Uri: !Sub arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${UploadLambdaFunction.Arn}/invocations
+        IntegrationResponses:
+          - ResponseParameters:
+              method.response.header.Access-Control-Allow-Headers: "'Content-Type,X-Amz-Date,Authorization,X-Api-Key,X-Amz-Security-Token'"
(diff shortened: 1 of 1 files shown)


stacks/upload.prx.org.yml
```diff
@@ -42,6 +42,18 @@ Resources:
         Type: AWS_PROXY
         IntegrationHttpMethod: POST
         Uri: !Sub arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${UploadLambdaFunction.Arn}/invocations
+        IntegrationResponses:
```
farski: It's hard for me to say if API Gateway will apply the headers without the template and model definitions. I would maybe add them just for completeness.

The other thing that needs to happen to make this all work is we need to make sure the CD deploy creates a new API Gateway _deployment_. The easiest way to do that is change the name of the key on L85 and update the `DeploymentId` ref on L98 to match.

pkarman: @farski ok I added the template, model and deployment id changes. PTAL.

stacks/upload.prx.org.yml
```diff
@@ -70,7 +88,7 @@ Resources:
             method.response.header.Access-Control-Allow-Methods: true
             method.response.header.Access-Control-Allow-Origin: true
           StatusCode: '200'
-  UploadRestApiDeployment:
+  UploadRestApiDeployment1.1:
```
farski: Resource ID's can only be `A-Za-z0-9`

pkarman: zapped the dot

PR #196, opened 2018-03-09
Dovetail stitch lambda logs/alarms

cavis:
🚧WIP StackSet to deploy lambda log-metrics/alarms in all our edge regions.

- [x] Pre-create the lambda log groups in all regions
- [x] CW log metrics to record cache hits/misses and errors
- [ ] ~CW alarms and notifications (to slack, probably)~


cavis:
Seems like we may be able to get away with only us-east-1 alarms, since we're doing all the stitching there.  And the CW 5XX alarm is doing a decent job.  Closing this - will open another ticket for alarms.

PR #197, opened 2018-03-09
Custom error TTLs

cavis:
Set explicit TTLs on 4xx / 5xx responses.

### DIFF
--- a/cdn/dovetail-cdn.yml
+++ b/cdn/dovetail-cdn.yml
@@ -11,9 +11,11 @@ Metadata:
       - Label:
           default: Distribution Parameters
         Parameters:
-          - OriginDomain
+          - OriginBucket
+          - OriginPath
           - DistributionDomain
-          - DistributionDomain
+          - CertificateArn
+          - LambdaFunctionArn
     ParameterLabels:
       OriginBucket:
         default: Origin S3 Bucket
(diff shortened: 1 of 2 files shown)


cdn/dovetail-cdn.yml
```diff
@@ -49,6 +49,27 @@ Resources:
       DistributionConfig:
         Aliases:
           - !Ref DistributionDomain
+        CustomErrorResponses:
+          # unrecognized uri, and s3 object not found
```
cavis: If a path doesn't look like `/<program>/<digest>/<filename>`, the lambda just passes that through.  Which triggers a 403 from S3 ... and this translates it back to a 404 before returning to the user.

cdn/dovetail-cdn.yml
```diff
@@ -49,6 +49,27 @@ Resources:
...
+          - ErrorCachingMinTTL: 300
+            ErrorCode: 403
+            ResponseCode: 404
+            ResponsePagePath: 404.html
+          # recognized uri, arrangement not in redis
```
cavis: It would be insane if a user got to a `<digest>` that wasn't in redis yet.  So the lambda explicitly returns a 404, and we set a 0-ttl on it, so it gets retried right away.

PR #198, opened 2018-03-12
Noisy origin-pull 403s

cavis:
Per PRX/dovetail-stitch-lambda#9 ... going to start returning 404s early.  And a 403 really means that the stitch-lambda uploaded a file, but the origin-pull couldn't access it (S3 eventual read consistency).

Also added the feeder `f.prxu.org` and `f-staging.prxu.org` CDNs into this PR.  Haven't switched the DNS from Highwinds yet, but they're in place when we're ready.

### DIFF
--- a/cdn/feeder-cdn.yml
+++ b/cdn/feeder-cdn.yml
@@ -62,7 +62,7 @@ Resources:
           SslSupportMethod: sni-only
       Tags:
         - Key: Project
-          Value: Dovetail
+          Value: feeder
         - Key: "prx:cloudformation:stack-name"
           Value: !Ref AWS::StackName
         - Key: "prx:cloudformation:stack-id"
--- a/cdn/feeder-cdn.yml
+++ b/cdn/feeder-cdn.yml
@@ -1,0 +1,76 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Description: >
+  Feeder bucket CDN (or really any plain S3-bucket CDN)
+Metadata:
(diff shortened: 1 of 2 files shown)


cavis:
@farski these are already deployed, if you want to take a look at this.

cdn/feeder-cdn.yml
```diff
@@ -0,0 +1,69 @@
...
+          - DomainName: !Ref OriginBucket
+            Id: feeder-s3-bucket
+            S3OriginConfig: {}
+        PriceClass: PriceClass_200
+        ViewerCertificate:
+          AcmCertificateArn: !Ref CertificateArn
+          SslSupportMethod: sni-only
+      Tags:
+        - Key: Project
+          Value: Dovetail
```
farski: This should probably be `feeder.prx.org`

cavis: Hmm... I think I pulled that tag from the source S3 bucket.  (`prx-feed.tag.Project == "Dovetail").  But I'm fine with feeder.  Should that be capitalized "Feeder" though?

farski: Oh hm. Maybe that should change too, if the bucket only has feeds from feeder. I'm not sure about `Feeder` vs `feeder.prx.org`. We're somewhat inconsistent between projects, and my phone is in the other room so I couldn't log in to check.

cavis: The ALBs have `Project = feeder` (lowercase) ... so I'll go with that!

PR #200, opened 2018-03-13
Edge lambda invoke

cavis:
Allow the stitch lambda to invoke an edge lambda of the same name.

For PRX/dovetail-stitch-lambda#10


cavis:
Tested in staging, and all the wildcards work: 

```
"Resource": "arn:aws:lambda:*:*:function:us-east-1.*DovetailStitchLambda*:*"
```

PR #205, opened 2018-03-26
adds template for RDS mediajoint instance

sandikbarr:
intended to be run manually as follows:
`aws cloudformation update-stack --stack-name <stack name> --template-url <template url> --parameters ParameterKey=DBName,ParameterValue=mediajoint_production ParameterKey=DBUser,ParameterValue=<admin but not actually SUPER user> ParameterKey=DBPassword,ParameterValue=<admin user password> ParameterKey=DBAllocatedStorage,ParameterValue=<GB of storage> ParameterKey=DBInstanceClass,ParameterValue=<RDS tier> ParameterKey=EC2SecurityGroup,ParameterValue=<ssh tunnel security group id> ParameterKey=RDSSubnetGroup,ParameterValue=<Platform-Production subnet group name>`

### DIFF
--- a/stacks/rds-mediajoint.yml
+++ b/stacks/rds-mediajoint.yml
@@ -1,0 +1,124 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Description: >-
+  AWS CloudFormation RDS Template: Template to create RDS MySQL v5.5 instance.
+Parameters:
+  DBName:
+    Default: my_database
+    Description: The database schema name
+    Type: String
+    MinLength: '1'
+    MaxLength: '64'
+    AllowedPattern: '[a-zA-Z][a-zA-Z0-9\-_]*'
+    ConstraintDescription: >-
+      must begin with a letter and contain only alphanumeric characters, dash,
+      or underscore.
+  DBUser:
+    NoEcho: 'true'
+    Description: The database admin account username
+    Type: String
+    MinLength: '1'
+    MaxLength: '16'
+    AllowedPattern: '[a-zA-Z][a-zA-Z0-9]*'
+    ConstraintDescription: must begin with a letter and contain only alphanumeric characters.
+  DBPassword:
+    NoEcho: 'true'
+    Description: The database admin account password
+    Type: String
+    MinLength: '1'
+    MaxLength: '41'
+    AllowedPattern: '[a-zA-Z0-9]+'
+    ConstraintDescription: must contain only alphanumeric characters.
+  DBAllocatedStorage:
+    Default: '20'
+    Description: The size of the database (Gb)
+    Type: Number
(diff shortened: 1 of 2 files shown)


stacks/rds-mediajoint.yml
```diff
@@ -0,0 +1,124 @@
+AWSTemplateFormatVersion: "2010-09-09"
```
farski: The `/stacks` directory has kind of stumbled into being reserved for stacks that are deployed though CD. I suggest we move this to a `/db` directory (to match `/dns` and `/cdn`, which also have stacks that are managed outside of CD)

stacks/rds-mediajoint.yml
```diff
@@ -0,0 +1,124 @@
...
+      StorageType: gp2
+      DBInstanceClass: !Ref DBInstanceClass
+      Engine: MySQL
+      EngineVersion: 5.5.46
+      MasterUsername: !Ref DBUser
+      MasterUserPassword: !Ref DBPassword
+      MultiAZ: !Ref MultiAZ
+      Tags:
+        - Key: Name
+          Value: Master Database
```
farski: There's a number of common tags that should get added to all taggable resources in templates. See _Tag everything_ [here](https://github.com/PRX/Infrastructure/blob/master/cd/README.md) and [an example](https://github.com/PRX/Infrastructure/blob/master/cdn/feed-cdn.yml#L156)

stacks/rds-mediajoint.yml
```diff
@@ -0,0 +1,124 @@
...
+      - db.t2.small
+      - db.t2.medium
+      - db.t2.large
+    ConstraintDescription: must select a valid database instance type.
+  EC2SecurityGroup:
+    Description: >-
+      The "MySQL_replication_tunnel" EC2 security group that contains the EC2 instance acting as the replication tunnel
+      https://console.aws.amazon.com/ec2/v2/home?region=us-east-1#SecurityGroups:sort=groupId
+    Default: default
+    Type: String
```
farski: Security group names & ID's and some other AWS related values can be set explicitly as the parameter type https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#aws-specific-parameter-types

farski:
I've had better luck with `aws cloudformation deploy` than `aws cloudformation update-stack`. The `deploy` behavior is a lot closer to what the console does, especially in terms of parameters. `update-stack`, I believe, requires all parameters get passed in for every call, where `deploy` will use existing parameter values if a replacement isn't provided.

sandikbarr:
Updated the stack with a `deploy` command. I can see where that's more convenient for small changes like the new tags and parameter type. Back to you @farski

PR #218, opened 2018-04-12
Proxy certs

cavis:
Have proxy just create its own certs.  And make sure those work for `https://prx.org` ... so it can redirect that to www.

### DIFF
--- a/stacks/proxy.prx.org.yml
+++ b/stacks/proxy.prx.org.yml
@@ -16,29 +16,51 @@ Parameters:
     Type: String
   EnvironmentTypeAbbreviation:
     Type: String
-  WildcardCertificateArn:
-    Type: String
 Mappings:
   EnvironmentTypeMap:
     Testing:
       DomainName: "proxy.test.prx.tech"
       HostedZoneName: "prx.tech."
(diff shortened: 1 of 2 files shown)


stacks/proxy.prx.org.yml
```diff
@@ -16,29 +16,51 @@ Parameters:
...
+      DomainName: !FindInMap [EnvironmentTypeMap, !Ref EnvironmentType, DomainName]
+      DomainValidationOptions:
+        - ValidationDomain: !FindInMap [EnvironmentTypeMap, !Ref EnvironmentType, ValidationDomain]
+          DomainName: !FindInMap [EnvironmentTypeMap, !Ref EnvironmentType, DomainName]
+      SubjectAlternativeNames:
+        - "prx.org"
+        - "www.prx.org"
+      Tags:
+        - Key: Project
+          Value: Corporate
```
farski: `proxy.prx.org`

cavis: should i change the tag on the lambda function as well then?

farski: Yes please

PR #220, opened 2018-04-16
Downsize prod cluster

cavis:
See PRX/dovetail.prx.org#259

Currently we run around 8 `m4.large` instances in prod.  This is less necessary now that DT stitching is _mostly_ done in lambdas.  This moves the cluster back to `t2.medium` and changes some of the memory reservations to accommodate more tasks per instance.

Will want to disable prod deploys while testing this out in staging.  To make sure the t2 unlimited stuff actually works.

### DIFF
--- a/stacks/ecs.yml
+++ b/stacks/ecs.yml
@@ -231,6 +231,15 @@ Resources:
                 Action:
                   - "ec2:DescribeTags"
                 Resource: "*"
+        - PolicyName: ModifyCreditSpecificationPolicy
+          PolicyDocument:
+            Version: "2012-10-17"
+            Statement:
+              - Effect: Allow
+                Action:
+                  - ec2:ModifyInstanceCreditSpecification
+                  - ec2:DescribeInstanceCreditSpecifications
+                Resource: "*"
       ManagedPolicyArns:
         - "arn:aws:iam::aws:policy/service-role/AmazonEC2ContainerServiceforEC2Role"
         - !Ref SecretsInstanceDecryptPolicyArn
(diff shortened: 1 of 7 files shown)


stacks/ecs.yml
```diff
@@ -401,13 +405,16 @@ Resources:
...
               command: "(crontab -uroot -l 2> /dev/null; echo '*/5 * * * * docker ps -qa | xargs docker inspect --format=\"{{ .State.Pid }}\" | xargs -IZ /sbin/fstrim /proc/Z/root/') | crontab -u root -"
-            # 04_set_newrelic_data:
-            #   command: "echo \"display_name: $(curl -ss http://169.254.169.254/latest/meta-data/instance-id)\" >> /etc/newrelic-infra.yml"
-            # 05_install_newrelic_infra:
-            #   command: |
-            #     sudo curl -o /etc/yum.repos.d/newrelic-infra.repo https://download.newrelic.com/infrastructure_agent/linux/yum/el/6/x86_64/newrelic-infra.repo
-            #     sudo yum -q makecache -y --disablerepo='*' --enablerepo='newrelic-infra'
-            #     sudo yum install newrelic-infra -y
+            04_set_t2_unlimited:
+              command: !Sub |
```
cavis: Hrmm... did i actually need a `!Sub` here?

farski: Seems like no

farski: it may actually bork because of the `${JSON}` if you keep the `!Sub`

PR #225, opened 2018-05-03
Stack to make an elasticache memcached cluster

kookster:
Used by exchange

### DIFF
--- a/db/mysql-instance.yml
+++ b/db/mysql-instance.yml
@@ -84,10 +84,24 @@ Parameters:
     AllowedValues:
       - Staging
       - Production
+  VPCStack:
+    Description: VPC stack to add this cache into
+    Type: String
+    Default: VPCStack
 Conditions:
   UseSnapshot: !Not [!Equals [!Ref DBSnapshot, '']]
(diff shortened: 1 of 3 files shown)


db/mysql-instance.yml
```diff
@@ -170,7 +184,7 @@ Resources:
         - Key: 'prx:cloudformation:stack-id'
           Value: !Ref AWS::StackId
       DBSubnetGroupName: !Sub Platform-${EnvironmentType}
```
sandikbarr: needs to be the RDSSubnetGroup from above

db/mysql-instance.yml
```diff
@@ -169,8 +183,8 @@ Resources:
           Value: !Ref AWS::StackName
         - Key: 'prx:cloudformation:stack-id'
           Value: !Ref AWS::StackId
-      DBSubnetGroupName: !Sub Platform-${EnvironmentType}
-# The MySQL_replication_tunnel security group is added and removed manually for use during migration     
+      DBSubnetGroupName: !Sub Platform-${EnvironmentType}-${AWS::StackName}-db
```
kookster: missed this before, better @sandikbarr ?

PR #227, opened 2018-05-10
Add stripe proxy

kookster:
- [x] Add in an nginx proxy server 
- [x] Add in an ALB to front it for SSL

### DIFF
--- a/stacks/legacy-support/stripe-proxy.yml
+++ b/stacks/legacy-support/stripe-proxy.yml
@@ -36,23 +36,20 @@ Parameters:
   # App ENV #################################################################
   EcrRegion:
     Type: String
-  ContainerMemory:
-    Type: String
-    Default: 400
-  ContainerCpu:
-    Type: String
-    Default: 128
-  HealthCheckPath:
-    Type: String
-    Default: /healthcheck
 Mappings:
   EnvironmentTypeMap:
     Testing:
       ProxyCount: 1
+      ContainerMemory: 400
+      ContainerCpu: 128
     Staging:
       ProxyCount: 1
+      ContainerMemory: 400
+      ContainerCpu: 128
     Production:
       ProxyCount: 1
+      ContainerMemory: 400
+      ContainerCpu: 128
 Resources:
   StripeProxyALBSecurityGroup:
     Type: "AWS::EC2::SecurityGroup"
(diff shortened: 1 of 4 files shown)


stacks/root.yml
```diff
@@ -925,6 +929,44 @@ Resources:
           Value: !Ref AWS::StackId
       TemplateURL: !Join ["", ["http://s3", !If [IsUsEast1, "", "-"], !If [IsUsEast1, "", !Ref "AWS::Region"], ".amazonaws.com/", "Fn::ImportValue": !Sub "${InfrastructureStorageStackName}-InfrastructureSourceBucket", "/", !Ref InfrastructureGitCommit, "/stacks/proxy.prx.org.yml"]]
       TimeoutInMinutes: "5"
+  FeederStack:
```
farski: Feeder stack?

farski: Also we shouldn't be adding any new application stacks directly to the root stack. They should all be getting added to nested stacks (like the serverless or static-sites one that already exist).

stacks/root.yml
```diff
@@ -123,6 +123,8 @@ Mappings:
       DovetailMemory: 400
```
farski: Nitpicky, but we're trying to follow the [seven rules](https://chris.beams.io/posts/git-commit/#seven-rules) for this repo (capitalize commit messages, etc)

kookster: Is the issue the commit message, or something else? (I've just read them they are not consistent)

stacks/root.yml
farski: Hm. Not really. `legacy-support`?

farski: I guess ideally it would be something that we expect to add other things to in the future, so we don't end up with lots of nested stacks that only have one thing in them. So maybe something more generic than that?

stacks/legacy-support/stripe-proxy.yml
```diff
@@ -0,0 +1,284 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Description: >
+  Docker web application with optional worker, with its own ALB
```
farski: Does this have an optional worker?

stacks/root.yml
```diff
@@ -203,6 +203,35 @@ Resources:
...
+        ECSServiceIAMRole: !GetAtt ECSClusterStack.Outputs.ECSServiceIAMRole
+        EcrRegion: !Ref EcrRegion
+      Tags:
+        - Key: "prx:cloudformation:stack-name"
+          Value: !Ref AWS::StackName
+        - Key: "prx:cloudformation:stack-id"
+          Value: !Ref AWS::StackId
+      TemplateURL: !Join ["", ["http://s3", !If [IsUsEast1, "", "-"], !If [IsUsEast1, "", !Ref "AWS::Region"], ".amazonaws.com/", "Fn::ImportValue": !Sub "${InfrastructureStorageStackName}-InfrastructureSourceBucket", "/", !Ref InfrastructureGitCommit, "/stacks/legacy-support/root.yml"]]
+      TimeoutInMinutes: "5"
+
```
farski: Whitespace

PR #230, opened 2018-05-31
Add assumable role for VPC peering

farski:
Allows the role the be assumed by anyone in the organization


stacks/root.yml
```diff
@@ -15,6 +15,8 @@ Conditions:
   IsStaging: !Equals [!Ref EnvironmentType, Staging]
   IsProduction: !Equals [!Ref EnvironmentType, Production]
 Parameters:
+  AwsOrganizationId:
```
cavis: A stack doesn't know its own organization-id?

PR #244, opened 2018-06-28
Add alarms for mysql replication

kookster:
Just figuring out how to write these, probably wrong.
Will try running them and see!


kookster:
ok, this works now, deployed to production

db/replication-alarms.yml
```diff
@@ -0,0 +1,97 @@
...
+    Description: Instance id (i.e. i-something) of the EC2 server
+    Type: String
+Resources:
+  ReplicationIoHaltedAlarm:
+    Type: "AWS::CloudWatch::Alarm"
+    Properties:
+      ActionsEnabled: true
+      AlarmName: !Sub "[Replication][${EnvironmentType}][SlaveIORunning] Not running"
+      AlarmActions: [!Ref OpsWarnSnsTopic]
+      InsufficientDataActions: [!Ref OpsWarnSnsTopic]
```
farski: We almost never use the short form array syntax in yaml (only because I almost never use it), so consistency would have this change to the other form. I prefer the long form because it diffs better.

db/replication-alarms.yml
```diff
@@ -0,0 +1,97 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Description: >
+  Creates alarms related to ssh tunnel remote server -> rds db replication
+Parameters:
+  OpsWarnSnsTopic:
+    Default: ''
```
farski: Prefer double quotes to single quotes

db/replication-alarms.yml
```diff
@@ -0,0 +1,97 @@
...
+    Type: String
+    AllowedValues:
+      - staging
+      - production
+  DatabaseName:
+    Default: 'mediajoint_production'
+    Description: Name of the database
+    Type: String
+  TunnelInstanceId:
+    Default: ''
```
farski: Quotes

db/replication-alarms.yml
```diff
@@ -0,0 +1,97 @@
...
+    Description: ops warn SNS Topic arn
+    Type: String
+  EnvironmentType:
+    Type: String
+    AllowedValues:
+      - staging
+      - production
+  DatabaseName:
+    Default: 'mediajoint_production'
```
farski: Unnecessary quotes

db/replication-alarms.yml
```diff
@@ -0,0 +1,97 @@
...
+          Value: !Ref EnvironmentType
+        - Name: InstanceID
+          Value: !Ref TunnelInstanceId
+      EvaluationPeriods: 2
+      MetricName: SlaveSQLRunning
+      Namespace: PRX/Data
+      Period: 60
+      Statistic: Average
+      Threshold: 1
+      TreatMissingData: notBreaching
```
farski: Missing data here may also indicate something not running, so maybe this should be switched?

db/replication-alarms.yml
```diff
@@ -0,0 +1,97 @@
...
+          Value: !Ref DatabaseName
+        - Name: Environment
+          Value: !Ref EnvironmentType
+        - Name: InstanceID
+          Value: !Ref TunnelInstanceId
+      EvaluationPeriods: 2
+      MetricName: SlaveSQLRunning
+      Namespace: PRX/Data
+      Period: 60
+      Statistic: Average
```
farski: What does the value of this metric represent? A boolean?

kookster: boolen, yes - this is a 1 or 0, a 1 if it is running, a 0 if it is not.

db/replication-alarms.yml
kookster: good point, I can switch that

PR #248, opened 2018-07-09
Stack for a redis cluster

cavis:
🚧 WIP


farski:
I'm closing this. If we find that we need a Redis template we can start over with three more years of Cfn knowledge.

Issue #240, opened 2018-06-27
Parameter MasterUsername should NOT exist with DBSnapshotIdentifier for Resources/MasterDB/Properties


farski:
In `db/mysql-instance.yml` the `DBSnapshotIdentifier` value is conditional, and it seems like in cases where it resolves to something other than `AWS::NoValue`, the `MasterUsername` and `MasterUserPassword` shouldn't exist, but at the moment they always exist. I think we may just need to make them inversely conditional to `DBSnapshotIdentifier`.

pkarman:
@farski is this in progress? I see it in the sprint but no assignee.

farski:
Not in progress by me. I haven't really done any of that DB work what I'd need to do a good amount of research to know what's going on here

pkarman:
AWS reporting as an error.

pkarman:
Maybe @sandikbarr will have some insight as to syntax improvements? Leaving unassigned for group effort.

PR #250, opened 2018-07-16
fix linting

sandikbarr:
* MasterUsername/MasterUserPassword linting fix by creating a separate restored db instance resource
* cleans up type linting, quotes, and other non string types
* also adds DependsOn for the RDSSubnetGroup to finish creation before the database is created

### DIFF
--- a/db/mysql-instance.yml
+++ b/db/mysql-instance.yml
@@ -152,6 +152,7 @@ Resources:
 # CloudFormation cannot update a stack when a custom-named resource requires replacing,
 # so when replacement is required (such as when restoring from snapshot),
 # use a new instance name such as oldname-restoredYYYYMMDD
+# Better yet to self manage tearing down of the existing resource, use a new stack with a different instance name
       DBInstanceIdentifier: !If [NamedInstance, !Ref DBInstance, !Ref "AWS::NoValue"]
       DBName: !If [CreateDB, !Ref CreateDBName, !Ref "AWS::NoValue"]
       AllocatedStorage: !Ref DBAllocatedStorage
--- a/db/mysql-instance.yml
+++ b/db/mysql-instance.yml
@@ -147,6 +147,7 @@ Resources:
             !Sub ${VPCStack}-Subnet2
   MasterDB:
     Type: "AWS::RDS::DBInstance"
+    DependsOn: RDSSubnetGroup
     Properties:
 # CloudFormation cannot update a stack when a custom-named resource requires replacing,
 # so when replacement is required (such as when restoring from snapshot),
--- a/db/mysql-instance.yml
+++ b/db/mysql-instance.yml
@@ -146,7 +146,6 @@ Resources:
         - Fn::ImportValue:
             !Sub ${VPCStack}-Subnet2
   MasterDB:
-    Condition: CreateDB
     Type: "AWS::RDS::DBInstance"
     Properties:
 # CloudFormation cannot update a stack when a custom-named resource requires replacing,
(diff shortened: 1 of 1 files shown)


sandikbarr:
🤔 not sure why this codebuild fails?

farski:
I've done some testing, and I haven't been able to create a situation where a stack update tries to recreate the DB, even without a stack policy, unless I alter the value of a property on the `DBInstance` resource that is listed as an _update requires replacement_ property. I added/updated/removed other resources in the template, and made _update requires no interruption_ changes to the DBInstance resource, and those stack updates have all succeeded without any cycling of the instance itself.

I also messed around with that `PubliclyAccessible` property. CloudFormation seemed happy to update the instance both after manually changing the value and changing it back (i.e., the resource and the stack in sync), and also at a time after manually changing it before changing it back (i.e., the stack and reality were out of sync). CloudFormation definitely still thinks that changing that value through the template would require a replacement, though I think that's probably a mistake and I'm going to file a support ticket to at least ask.

As far as the linter error in #240, I ran an example through that included a `DBSnapshotIdentifier` and then explicitly set the value of `MasterUsername` and `MasterUserPassword` to `AWS::NoValue`. CloudFormation treats properties with `NoValue` as though those properties don't exist, so this actually seems like a bug with the linter; it should also ignore `NoValue` properties (I've filed a ticket). So even my suggestion in #240 wouldn't help here. This isn't the only bogus linter error that is keeping our templates from being error free so I'd be okay with doing that way and just ignoring the error for now.

So with all that in mind, what I think make sense here is to go back to a template with a single resource that conditionally is either a virgin DB or a snapshot restoration, and just make sure `MasterUsername` and `MasterUserPassword` get set to `AWS::NoValue` when necessary (even though it seems like we've been able to do snapshot restores with them there. The API docs definitely say they shouldn't be included, but I guess either RDS is ignoring them, or Cfn is omitting them). That will avoid the situation where a single stack launched from this template has two databases for a time. Instead, in cases where we do need to do a restore, that's entirely encapsulated by a new, separate stack just for the "restored" DB. At that point, we can tear down the old, original DB stack at whatever point it's no longer needed. Those two cases will just use the same template; similar to how to use the CDN template for a number of different types of CloudFront distributions.

I know you've spent more time looking at this than I have, so if any of my test results run counter to what you were seeing we should dig into that, just so we can be sure we know how the system is going to behave.

sandikbarr:
> I ran an example through that included a DBSnapshotIdentifier and then explicitly set the value of MasterUsername and MasterUserPassword to AWS::NoValue. CloudFormation treats properties with NoValue as though those properties don't exist, so this actually seems like a bug with the linter; it should also ignore NoValue properties (I've filed a ticket).

so this was a non-issue since those values are AWS::NoValue prior to this PR unless the database is being created?

> CloudFormation definitely still thinks that changing that value through the template would require a replacement

yup

> So with all that in mind, what I think make sense here is to go back to a template with a single resource that conditionally is either a virgin DB or a snapshot restoration, and just make sure MasterUsername and MasterUserPassword get set to AWS::NoValue when necessary

It's doing that now. As a separate concern though, I feel like it even may warrant a separate template for the restored database. Then we migrate and tear down the existing stack when ready.

So just close this PR? or were there further things you needed me to address with this bug in the linter?

farski:
>so this was a non-issue since those values are AWS::NoValue prior to this PR unless the database is being created?

From how I'm reading the pre-PR template, the values for `MasterUsername` and 
`MasterUserPassword` are never being set to `AWS::NoValue`.

sandikbarr:
sorry you're right that's just how I remembered it

Funny that this AWS example contradicts the whole linting error with username, password, and snapshot being the only params 🤔 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-sample-templates.html#w2ab2c21c28c21c47b5

farski:
Apparently they do that [a lot](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html#cfn-pseudo-param-novalue)

db/mysql-instance.yml
```diff
@@ -171,17 +172,25 @@ Resources:
         - Ref: AWS::NoValue
       Engine: MySQL
       EngineVersion: 5.5.46
-      MasterUsername: !Ref DBUser
-      MasterUserPassword: !Ref DBPassword
+      MasterUsername:
+        Fn::If:
```
farski: Let's switch these to the short form `If` to be consistent with our other templates and then I think this should be good for now (I think there's five total)

db/mysql-instance.yml
```diff
@@ -146,42 +146,31 @@ Resources:
         - Fn::ImportValue:
             !Sub ${VPCStack}-Subnet2
   MasterDB:
-    Type: 'AWS::RDS::DBInstance'
+    Type: "AWS::RDS::DBInstance"
+    DependsOn: RDSSubnetGroup
     Properties:
 # CloudFormation cannot update a stack when a custom-named resource requires replacing,
```
farski: I guess we also should remove this comment, since we don't really want people to use a preexisting stack to do snapshot recoveries

PR #254, opened 2018-07-27
Dovetail CDN bytes logging

cavis:
Stack for getting dovetail cloudfront-downloaded bytes into kinesis.


stacks/root.yml
```diff
@@ -837,6 +839,24 @@ Resources:
           Value: !Ref AWS::StackId
       TemplateURL: !Join ["", ["http://s3", !If [IsUsEast1, "", "-"], !If [IsUsEast1, "", !Ref "AWS::Region"], ".amazonaws.com/", "Fn::ImportValue": !Sub "${InfrastructureStorageStackName}-InfrastructureSourceBucket", "/", !Ref InfrastructureGitCommit, "/stacks/dovetail-stitch-lambda.yml"]]
       TimeoutInMinutes: 5
+  DovetailBytesLambdaStack:
```
farski: Can we find a nested stack to put this in, to keep the root template for getting any bigger?

cavis: maybe serverless?

PR #255, opened 2018-08-06
Stitch lambda updates

cavis:
- [x] Add the s3-tagging permission for the dovetail-stitch-lambda, which will be required for PRX/dovetail-stitch-lambda#28
- [x] Restrict stitch-lambda S3 writes to a subfolder (just in case)
- [x] Quiet alarms, by moving all the prod `> 0` to ops-warn, and adding new `> 2` prod alarms to ops-error

### DIFF
--- a/cdn/dovetail-cdn/dovetail-stitch-lambda-filters.yml
+++ b/cdn/dovetail-cdn/dovetail-stitch-lambda-filters.yml
@@ -131,26 +131,14 @@ Resources:
       ActionsEnabled: true
       AlarmName: !Sub "[DovetailStitchLambda][Errors] ${EnvironmentType} > 0"
       AlarmActions:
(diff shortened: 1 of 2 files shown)


cdn/dovetail-cdn/dovetail-stitch-lambda-filters.yml
```diff
@@ -164,6 +152,34 @@ Resources:
       Dimensions:
         - Name: FunctionName
           Value: !Sub "${LambdaRegion}.${LambdaName}"
+  LambdaManyErrorsAlarm:
+    Type: "AWS::CloudWatch::Alarm"
+    Condition: !And [CreateLambdaAlarms, IsProduction]
```
cavis: Is this going to work?

PR #256, opened 2018-08-31
Secrets lambda ignores files other than secrets

kookster:
Enhancing `aws-secrets` to securely store files other than just `/app/env/secrets` (e.g. private key files). https://github.com/PRX/aws-secrets/pull/6

However, if they are stored in the same `/app/env/*` path, they are picked up as changes to the secrets.

This change causes the secrets update lambda to ignore changes to files that are not names `/secrets`

### DIFF
--- a/secrets/README.md
+++ b/secrets/README.md
@@ -2,7 +2,8 @@
 
 Works with the `aws-secrets` project (https://github.com/PRX/aws-secrets) to
 provide the bucket to store secrets, and creates a lambda to watch for changes
-to secrets files in that S3 bucket.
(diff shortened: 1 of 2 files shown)


kookster:
@farski is there a good way to write tests for these? hmm

farski:
@kookster Yeah you could use unittest or really whatever you want and just add them to the `buildspec`

farski:
@kookster Not merging in case this still needs to get deployed; didn't want it to get lost.

kookster:
this is now deployed, so merging and closing

PR #260, opened 2018-09-10
Improve CodeBuild callback handling

### LINKED ISSUE #157: CodeBuild Callback Slack messages are not quite right
The handling of the text value for successful builds only supports a pull request OR ecr OR lambda. Theoretically it could be any number of those, so the message should be able to include info about all of them at once

farski:
Fixes #157


farski:
Already deployed

PR #264, opened 2018-10-03
Dovetail counts lambda

cavis:
Requires PRX/dovetail-counts-lambda#3 first.  I'm not completely confident all of these things will work!

### DIFF
--- a/stacks/serverless/dovetail-counts-lambda.yml
+++ b/stacks/serverless/dovetail-counts-lambda.yml
@@ -14,8 +14,8 @@ Parameters:
     Type: String
   EnvironmentTypeAbbreviation:
     Type: String
-  VPC:
-    Type: "AWS::EC2::VPC::Id"
+  VPCSecurityGroup:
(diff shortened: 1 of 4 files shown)


stacks/serverless/dovetail-counts-lambda.yml
```diff
@@ -0,0 +1,108 @@
...
+          REDIS_URL: !Ref RedisUrl
+          S3_BUCKET: !Ref S3Bucket
+          S3_PREFIX: !Ref S3Prefix
+      Handler: index.handler
+      MemorySize: 128
+      Role: !GetAtt DovetailCountsLambdaExecutionIAMRole.Arn
+      Runtime: nodejs8.10
+      Timeout: 10
+      VpcConfig:
+        SecurityGroupIds: !GetAtt VPC.DefaultSecurityGroup
```
cavis: Found some forum post with something like this in there ... but not sure it'll actually work.

PR #267, opened 2018-10-08
Add a CNAME from for api.prx.org to exchange tech hosted on aws

kookster:
`api.prx.org` has always gone to the exchange app, since it was hosted on contegix at the apex domain.
Need to specify this explicitly now to go to AWS via the `exchange.prx.tech` address.

### DIFF
--- a/dns/prx.org-hosted_zone.yml
+++ b/dns/prx.org-hosted_zone.yml
@@ -372,11 +372,6 @@ Resources:
           TTL: "300"
           Type: CNAME
           Name: !Sub exchange.${Domain}
-        - ResourceRecords:
-            - exchange.prx.tech.
-          TTL: "300"
-          Type: CNAME
-          Name: !Sub api.${Domain}
         - Type: A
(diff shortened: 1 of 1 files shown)


cavis:
@kookster this is deployed, right?

farski:
@cavis it's deployed, but the record is still in the `Prx01` group, which is no longer accurate.

There are three options:
1. Ignore the mismatch
2. Do one deploy without the record, and then another deploy with the record in the correct group, with the understanding that there could be as much as 5 minutes where `api.prx.org` is inaccessible (though more than likely there would be no downtime)
3. Deploy a new stack with the record in the correct group, and swap the name servers on the domain for those belonging to the newly-created hosted zone, to avoid any downtime.

I'd vote for 2.

cavis:
Yeah, I think (2) was what i did with `www` way back when.  And this actually has less impact than that.  Will do!

cavis:
Deploy'd!

PR #272, opened 2018-10-16
CMS audio lambda

cavis:
For PRX/internal#285.

Does not add the SNS subscriptions for the Lambda.  I'll do that manually to make sure everything is working, then CFN-ize that subscription and nuke the old Lambdas.


stacks/serverless/cms-audio-lambda.yml
```diff
@@ -0,0 +1,184 @@
+# stacks/serverless/cms-audio-lambda.yml
+AWSTemplateFormatVersion: "2010-09-09"
+Description: Audio processing for cms.prx.org
+Parameters:
+  AlarmWarnActions:
+    Type: CommaDelimitedList
```
cavis: This may or may not work.  But simplifies this template to put the ops-warn vs ops-error logic in the root template.

stacks/serverless/root.yml
```diff
@@ -134,3 +134,35 @@ Resources:
...
       TemplateURL: !Join ["", [!Ref TemplateUrlPrefix, "dovetail-counts-lambda.yml"]]
       TimeoutInMinutes: 5
+  CmsAudioLambdaStack:
+    Type: "AWS::CloudFormation::Stack"
+    Properties:
+      NotificationARNs:
+        - Fn::ImportValue:
+            !Sub "${InfrastructureNotificationsStackName}-CloudFormationNotificationSnsTopic"
+      Parameters:
+        AlarmWarnActions:
```
cavis: This passes in a String as the parameter ... but the `CommaDelimitedList` should transform it into a 1-item list.

PR #274, opened 2018-10-30
Feat/clean up template urls

farski:
This moves a lot of the logic for building the nested stack template URLs into the storage stack, and exports a complete URL that can be imported in the CD templates.

It also switched to https URLs, rather than http, which the docs lead me to believe should be fine.

CloudFormation does **not** treat a template URL changing as a change to any resources, unless the contents of what lives at the URL have in fact changed.

### DIFF
--- a/stacks/root.yml
+++ b/stacks/root.yml
@@ -426,7 +426,11 @@ Resources:
           Value: !Ref AWS::StackName
         - Key: "prx:cloudformation:stack-id"
           Value: !Ref AWS::StackId
-      TemplateURL: !Join ["", ["http://s3", !If [IsUsEast1, "", "-"], !If [IsUsEast1, "", !Ref "AWS::Region"], ".amazonaws.com/", "Fn::ImportValue": !Sub "${InfrastructureStorageStackName}-InfrastructureSourceBucket", "/", !Ref InfrastructureGitCommit, "/stacks/fourohfour.prx.org.yml"]]
+      TemplateURL: !Join
+        - ""
+        - - Fn::ImportValue: { "Fn::Sub": "${InfrastructureStorageStackName}-SourceBucketUrl" }
+          - !Ref InfrastructureGitCommit
+          - "/stacks/fourohfour.prx.org.yml"
       TimeoutInMinutes: 5
   PublishStack:
     Type: "AWS::CloudFormation::Stack"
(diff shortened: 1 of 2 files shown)


farski:
This moves a lot of the logic for building the nested stack template URLs into the storage stack, and exports a complete URL that can be imported in the CD templates. 

It also switched to `https` URLs, rather than `http`, which the docs lead me to believe should be fine.

CloudFormation does **not** treat a template URL changing as a change to any resources, unless the contents of what lives at the URL have in fact changed.

stacks/root.yml
```diff
@@ -426,7 +426,11 @@ Resources:
           Value: !Ref AWS::StackName
         - Key: "prx:cloudformation:stack-id"
           Value: !Ref AWS::StackId
-      TemplateURL: !Join ["", ["http://s3", !If [IsUsEast1, "", "-"], !If [IsUsEast1, "", !Ref "AWS::Region"], ".amazonaws.com/", "Fn::ImportValue": !Sub "${InfrastructureStorageStackName}-InfrastructureSourceBucket", "/", !Ref InfrastructureGitCommit, "/stacks/fourohfour.prx.org.yml"]]
+      TemplateURL: !Join
+        - ""
+        - - Fn::ImportValue: { "Fn::Sub": "${InfrastructureStorageStackName}-SourceBucketUrl" }
```
farski: I _think_ doing this on one line like this should work.

storage/storage.yml
```diff
@@ -87,6 +89,23 @@ Outputs:
...
+        - - "https://s3"
+          - !If
+            - IsUsEast1
+            - ""
+            - !Sub "-${AWS::Region}"
+          - ".amazonaws.com/"
+          - !Ref InfrastructureSourceBucket
+          - "/"
+    Export:
+      Name: !Sub ${AWS::StackName}-SourceBucketUrl
```
farski: This has already been deployed

stacks/root.yml
```diff
@@ -426,7 +426,11 @@ Resources:
           Value: !Ref AWS::StackName
         - Key: "prx:cloudformation:stack-id"
           Value: !Ref AWS::StackId
-      TemplateURL: !Join ["", ["http://s3", !If [IsUsEast1, "", "-"], !If [IsUsEast1, "", !Ref "AWS::Region"], ".amazonaws.com/", "Fn::ImportValue": !Sub "${InfrastructureStorageStackName}-InfrastructureSourceBucket", "/", !Ref InfrastructureGitCommit, "/stacks/fourohfour.prx.org.yml"]]
```
farski: I'm not sad to see this disaster going away

PR #281, opened 2018-11-28
notify slack when new series are created

svevang:
This PR sets up a new lambda that monitors the CMS create SNS topic and posts details to slack.

### DIFF
--- a/etc/cms-resource-reporter.yml
+++ b/etc/cms-resource-reporter.yml
@@ -1,0 +1,120 @@
+# etc/cms-resource-reporter.yml
+AWSTemplateFormatVersion: "2010-09-09"
+Description: Handles CMS Resource SNS Events
+Parameters:
+  SlackMessageRelayTopicArn:
+    Type: String
+  CmsSeriesCreateSnsTopicArn:
+    Type: String
+Resources:
+  LambdaIamRole:
+    Type: "AWS::IAM::Role"
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+            - Effect: "Allow"
+              Principal:
+                Service:
+                  - "lambda.amazonaws.com"
+              Action:
+                - "sts:AssumeRole"
+      Path: "/"
+      Policies:
+        - PolicyName: SnsPolicy
+          PolicyDocument:
+            Version: "2012-10-17"
+            Statement:
+              - Effect: Allow
+                Action:
+                  - "sns:Publish"
+                Resource:
+                  - !Ref SlackMessageRelayTopicArn
+      ManagedPolicyArns:
+        - "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
+  PushSubscription:
+    Type: "AWS::SNS::Subscription"
+    Properties:
+      Endpoint: !GetAtt MessageHandlerFunction.Arn
+      Protocol: "lambda"
+      Region: "us-east-1"
+      TopicArn: !Ref CmsSeriesCreateSnsTopicArn
+  LambdaResourcePolicy:
+    Type: "AWS::Lambda::Permission"
+    Properties:
+      FunctionName: !GetAtt MessageHandlerFunction.Arn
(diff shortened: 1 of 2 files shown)


etc/cms-resource-reporter.yml
```diff
@@ -0,0 +1,120 @@
...
+              };
+
+              var sns = new AWS.SNS({region: 'us-east-1'});
+              var relayParams = {
+                  Message: JSON.stringify(relayParams), 
+                  Subject: "CMS Resource Reporter: Resource Action Discovered",
+                  TopicArn: process.env.SLACK_MESSAGE_RELAY_SNS_TOPIC_ARN
+              };
+
+              return new Promise(function(resolve, reject) {
```
cavis: Since you're using async/await already ... maybe rewrite as:

```
var response = await sns.publish(relayParams).promise()
return relayParams
```

etc/cms-resource-reporter.yml
```diff
@@ -0,0 +1,120 @@
...
+      Handler: index.handler
+      MemorySize: 128
+      Role: !GetAtt LambdaIamRole.Arn
+      Runtime: nodejs8.10
+      Timeout: 8
+  LambdaErrorAlarm:
+    Type: "AWS::CloudWatch::Alarm"
+    Properties:
+      ActionsEnabled: true
+      AlarmName: "[CMSResourceReporter][Lambda][Error] Invocation 4XX"
```
cavis: 4XX?

etc/cms-resource-reporter.yml
```diff
@@ -0,0 +1,120 @@
...
+              var message = JSON.parse(event["Records"][0]["Sns"]["Message"]);
+              var event = JSON.parse(message["body"]);
+
+              var resourceLink = event['_links']['self']['href'];
+              var fromImport = event['_links']['prx:podcast-imports']['count'] > 0;
+
+
+              var importLabel = fromImport ? "import" : "publish"
+              var eventText = `A new series was created on CMS via *${importLabel}*:
+          _${event['title']}_
```
cavis: I like the accurate labeling.  Is this multiline thing necessary to make the slack formatting look nice?

etc/cms-resource-reporter.yml
```diff
@@ -0,0 +1,120 @@
+# etc/cms-resource-reporter.yml
+AWSTemplateFormatVersion: "2010-09-09"
+Description: Handles CMS Resource SNS Events
+Parameters:
+  SlackMessageRelayTopicArn:
+    Type: String
+  CmsSeriesCreateSnsTopicArn:
```
cavis: Can this stack be generalized to monitor any announce arn?  Maybe using the SNS topic parameter (create/update/delete), plus a few labels/nouns to describe it ("was created" / "was updated").  Or am I over thinking this.

svevang: Took a shot at an abstract resource reporter in 7a1990b.  It ends up being a less nicely formatted, but I think still very usable/practical.

etc/resource-reporter.yml
```diff
@@ -0,0 +1,121 @@
...
+      Handler: index.handler
+      MemorySize: 128
+      Role: !GetAtt LambdaIamRole.Arn
+      Runtime: nodejs8.10
+      Timeout: 8
+  LambdaErrorAlarm:
+    Type: "AWS::CloudWatch::Alarm"
+    Properties:
+      ActionsEnabled: true
+      AlarmName: "[CMSResourceReporter][Lambda][Error] Invocation Error"
```
cavis: I think these alarm names have to be unique in the account.  So trying to create more than one of these is going to fail.  Wondering if some sort of "human readable" input parameter could suffice?  Like `reportName = "Series Created"` or something?

cavis:
Odd that CodeBuild failed for this.  Looks like an ssl problem ... the `cfn-lint` actually worked.

etc/resource-reporter.yml
svevang: Thanks, I didn't know that about a cloudwatch alarms.  I added a new parameter to let users add a unique name.

PR #286, opened 2018-12-09
Add release notes prompt to deploy approvals

### LINKED ISSUE #275: Add release notes reminder to deploy button
[Here](https://github.com/PRX/Infrastructure/blob/master/notifications/lambdas/ike-interactive-messages-callback-handler/index.js#L143) would be a good spot to make the [`chat.postEphemeral`](https://api.slack.com/methods/chat.postEphemeral) call.

Currently that lambda doesn't make any Slack API calls; it's just handling a request from Slack, and returning an empty response. So we'll need to get a Slack token from somewhere to actually be able to make that call. Such a token should already exist somewhere, since we're sending messages from other Lambdas.

Alternatively there may be a way to use the slack message relay, and just send an object to that SNS topic so this lambda doesn't need to have its own ability to hit the Slack API. I can't remember if the Incoming Webhook api allows you to make ephemeral messages, and that's ultimately what the relay uses.

farski:
Add an additional deployment approval button that will, in addition to approving a production deploy, open a dialog prompting for release notes, which will get posted to #tech-releases

This is not really tested

Closes #275

### DIFF
--- a/cd/lambdas/pipeline-approval-notification-handler/lambda_function.py
+++ b/cd/lambdas/pipeline-approval-notification-handler/lambda_function.py
@@ -6,6 +6,10 @@
 # deploy. Information regarding the pending changes to the stack are queried
 # from CloudFormation, and sent to Slack along with control to make a decision
 # about the deploy.
+#
+# This creates the Deploy button for production deploys, but does NOT handle
+# what happens when the button is pressed. That is handled by
+# notifications/ike-interactive-messages-callback-handler.
 
 import boto3
 import json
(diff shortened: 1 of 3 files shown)


cd/lambdas/pipeline-approval-notification-handler/lambda_function.py
```diff
@@ -143,6 +147,17 @@ def approval_action_attachment(notification):
                     'ok_text': 'Deploy',
                     'dismiss_text': 'Cancel'
                 }
+            }, {
+                'type': 'button',
```
cavis: Why 2 buttons?  Just to make the non-release-notes deploys less annoying?

notifications/lambdas/ike-interactive-messages-callback-handler/index.js
```diff
@@ -168,8 +253,35 @@ function handleCodePipelineApproval(payload, callback) {
...
+                    trigger_id: payload.trigger_id,
+                    token: process.env.SLACK_ACCESS_TOKEN,
+                    dialog: JSON.stringify({
+                        callback_id: RELEASE_NOTES_DIALOG_CALLBACK,
+                        state: action.value,
+                        title: 'Release Notes',
+                        submit_label: 'Post',
+                        elements: [
+                            {
+                                type: 'textarea',
```
cavis: Ooh, fancy.

notifications/lambdas/ike-interactive-messages-callback-handler/index.js
```diff
@@ -52,18 +110,38 @@ function processEvent(event, context, callback) {
...
     }
 }
 
+function handleReleaseNotesDialog(payload, callback) {
+    console.log(JSON.stringify(payload));
+
+    sns.publish({
+        TopicArn: process.env.SLACK_MESSAGE_RELAY_TOPIC_ARN,
+        Message: JSON.stringify({
+            channel: '#tech-releses',
```
farski: Well that's spelt wrong

PR #294, opened 2019-01-29
Add remix application with scheduled task

kookster:
Create a template for adding CloudWatch event scheduled triggers for tasks
Have the web-application.yml export task definitions to schedule
Add the remix application
Add a scheduled task for remix with hourly command to `load_episodes`

### DIFF
--- a/stacks/container-apps/root.yml
+++ b/stacks/container-apps/root.yml
@@ -76,6 +76,13 @@ Parameters:
     Type: String
   CourierEcrImageTag:
     Type: String
+  # Remix Specific ###############################################################
+  RemixPlatformALBListenerPriorityPrefix:
+    Type: String
+  RemixSecretsVersion:
+    Type: String
+  RemixEcrImageTag:
+    Type: String
 Resources:
   ExchangeStack:
     Type: "AWS::CloudFormation::Stack"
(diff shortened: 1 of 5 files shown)


stacks/container-apps/root.yml
```diff
@@ -281,3 +288,70 @@ Resources:
...
+        SecretsBase: !Ref SecretsBase
+        ContainerMemory: !Ref ContainerMemory
+        ContainerMemoryReservation: !Ref ContainerMemoryReservation
+        ContainerCpu: !Ref ContainerCpu
+        PlatformALBListenerPriorityPrefix: !Ref RemixPlatformALBListenerPriorityPrefix
+        EcrImageTag: !Ref RemixEcrImageTag
+        SecretsVersion: !Ref RemixSecretsVersion
+        AppName: "remix"
+        CreateWorker: "true"
+        DesiredWorkersProduction: "0"
```
kookster: We don't want a worker running all the time, but I do want the task def to use for scheduling, so having it created, and set to 0 running

stacks/container-apps/root.yml
```diff
@@ -281,3 +288,70 @@ Resources:
...
+        ScheduleCommand: "[\"load_episodes\"]"
+        ScheduleExpression: "rate(1 hour)"
+        ScheduleTaskDefinitionName: "remix-worker"
+        ScheduleTaskDefinitionArn: !GetAtt RemixStack.Outputs.WorkerTaskDefinition
+      Tags:
+        - Key: "prx:cloudformation:stack-name"
+          Value: !Ref AWS::StackName
+        - Key: "prx:cloudformation:stack-id"
+          Value: !Ref AWS::StackId
+      TemplateURL: !Join ["", [!Ref TemplateUrlPrefix, "scheduled-task.yml"]]
```
kookster: new template!

stacks/container-apps/root.yml
```diff
@@ -281,3 +288,70 @@ Resources:
...
+      NotificationARNs:
+        - Fn::ImportValue:
+            !Sub "${InfrastructureNotificationsStackName}-CloudFormationNotificationSnsTopic"
+      Parameters:
+        ECSCluster: !Ref ECSCluster
+        ECSServiceIAMRole: !Ref ECSServiceIAMRole
+        ScheduleCommand: "[\"load_episodes\"]"
+        ScheduleExpression: "rate(1 hour)"
+        ScheduleTaskDefinitionName: "remix-worker"
+        ScheduleTaskDefinitionArn: !GetAtt RemixStack.Outputs.WorkerTaskDefinition
```
kookster: I believe this is the right syntax to reference the `Outputs` from the prior task

stacks/container-apps/root.yml
cavis: So you'll actually end up with a worker service then, right?  (But the desired count set to 0).

stacks/container-apps/root.yml
```diff
@@ -281,3 +288,70 @@ Resources:
...
+        ContainerCpu: !Ref ContainerCpu
+        PlatformALBListenerPriorityPrefix: !Ref RemixPlatformALBListenerPriorityPrefix
+        EcrImageTag: !Ref RemixEcrImageTag
+        SecretsVersion: !Ref RemixSecretsVersion
+        AppName: "remix"
+        CreateWorker: "true"
+        DesiredWorkersProduction: "0"
+        DesiredWorkers: "0"
+        ContainerPort: "3000"
+        HealthCheckPath: "/api"
```
cavis: Didn't realize this had an api!

kookster: it does! originally for the remix native mobile apps

stacks/container-apps/root.yml
kookster: yes, there will be a service there, that won't be run, but we'll use the task def

stacks/container-apps/root.yml
```diff
@@ -281,3 +288,70 @@ Resources:
...
+    Properties:
+      NotificationARNs:
+        - Fn::ImportValue:
+            !Sub "${InfrastructureNotificationsStackName}-CloudFormationNotificationSnsTopic"
+      Parameters:
+        ECSCluster: !Ref ECSCluster
+        ECSServiceIAMRole: !Ref ECSServiceIAMRole
+        ScheduleCommand: "[\"load_episodes\"]"
+        ScheduleExpression: "rate(1 hour)"
+        ScheduleTaskDefinitionName: "remix-worker"
```
farski: Where does this come from? It seems like it needs to exist for the task def. before the `containerOverrides`, but it only looks like it exists in the context of the events rule

stacks/container-apps/root.yml
```diff
@@ -281,3 +288,70 @@ Resources:
...
+    Condition: IsStaging
+    Type: "AWS::CloudFormation::Stack"
+    Properties:
+      NotificationARNs:
+        - Fn::ImportValue:
+            !Sub "${InfrastructureNotificationsStackName}-CloudFormationNotificationSnsTopic"
+      Parameters:
+        ECSCluster: !Ref ECSCluster
+        ECSServiceIAMRole: !Ref ECSServiceIAMRole
+        ScheduleCommand: "[\"load_episodes\"]"
```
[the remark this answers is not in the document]
farski: If you wrap the yaml string in single quotes you should be able to get away without the escaping

stacks/container-apps/root.yml
kookster: Added a `DependsOn` to make this more explicit

stacks/container-apps/root.yml
```diff
@@ -281,3 +288,71 @@ Resources:
...
+        HealthCheckPath: "/api"
+      Tags:
+        - Key: "prx:cloudformation:stack-name"
+          Value: !Ref AWS::StackName
+        - Key: "prx:cloudformation:stack-id"
+          Value: !Ref AWS::StackId
+      TemplateURL: !Join ["", [!Ref TemplateUrlPrefix, "web-application.yml"]]
+      TimeoutInMinutes: 5
+  RemixScheduledTaskStack:
+    DependsOn: RemixStack
```
kookster: @farski clearer?

stacks/container-apps/root.yml
farski: Oh I see, the `AppName` is used to derive that in `web-application.yml`. I think what I would do is, export either the AppName from the app stack, or export the entire worker name from the app stack (just as a sub that matches the one that makes the name in the task def), and then `GetAtt` that like you do for the Arn. I'd also get rid of the `DependsOn`; those shouldn't get used unless we explicitly need to control the order Cfn does things.

kookster:
linting errors I see in the log are unrelated:
```
E3030 You must specify a valid value for ComparisonOperator (LessThanOrEqualToThreshold).
Valid values are [u'GreaterThanOrEqualToThreshold', u'GreaterThanThreshold', u'LessThanOrEqualToThreshold.', u'LessThanThreshold']
./db/mysql-instance.yml:196:7

E3030 You must specify a valid value for ComparisonOperator (LessThanOrEqualToThreshold).
Valid values are [u'GreaterThanOrEqualToThreshold', u'GreaterThanThreshold', u'LessThanOrEqualToThreshold.', u'LessThanThreshold']
./stacks/dovetail.prx.org.yml:441:7
```

kookster:
looks like a problem in the linter to me, it is looking for "u'LessThanOrEqualToThreshold.' with a period at the end, and calling `LessThanOrEqualToThreshold` invalid

kookster:
bug in the linter has been fixed.

kookster:
ok - this looks ready to go to me - I added the missing root params for remix - merge time? @cavis

cavis:
👍 looks good

PR #297, opened 2019-02-07
Create a new postgresql template

svevang:
Based on a pared-down version of the MySQL template, this PR adds a new CFN template that creates RDS PostgreSQL resources.

Per:
https://github.com/PRX/courier.prx.org/issues/8
https://github.com/PRX/augury.prx.org/issues/1

### DIFF
--- a/db/postgresql-instance.yml
+++ b/db/postgresql-instance.yml
@@ -1,0 +1,219 @@
+# db/rds-instance.yml
+AWSTemplateFormatVersion: "2010-09-09"
+Description: >-
+  AWS CloudFormation RDS Template: Template to create a PostgreSQL RDS instance.
+Parameters:
+  DBInstance:
+    Default: "name-me"
+    Description: >-
+      The database instance name.
+      When replacement is required, must be different than existing instance name.
+    Type: String
+    MinLength: 0
+    MaxLength: 64
+    AllowedPattern: ^$|[a-zA-Z][a-zA-Z0-9\-]*
+    ConstraintDescription: >-
+      optional, must begin with a letter and contain only alphanumeric characters or dash.
+  CreateDBName:
+    Default: ""
+    Description: The database name for create only
+    Type: String
+    MinLength: 0
+    MaxLength: 64
+    AllowedPattern: ^$|[a-zA-Z][a-zA-Z0-9\-_]*
+    ConstraintDescription: >-
+      must begin with a letter and contain only alphanumeric characters, dash,
+      or underscore.
+  DBUser:
+    NoEcho: true
+    Description: The database admin account username
+    Type: String
+    MinLength: 1
+    MaxLength: 16
+    AllowedPattern: ^[a-zA-Z][a-zA-Z0-9_]*$
+    ConstraintDescription: must begin with a letter and contain only alphanumeric characters.
+  DBPassword:
+    NoEcho: true
+    Description: The database admin account password
+    Type: String
+    MinLength: 1
+    MaxLength: 41
+    AllowedPattern: ^[a-zA-Z0-9]+$
+    ConstraintDescription: must contain only alphanumeric characters.
+  DBAllocatedStorage:
+    Default: 20
+    Description: The size of the database (Gb)
+    Type: Number
+    MinValue: 20
+    MaxValue: 16384
+    ConstraintDescription: must be between 5 and 16384Gb.
+  DBInstanceClass:
+    Description: The database instance type
+    Type: String
+    Default: db.t2.small
+    ConstraintDescription: must select a valid database instance type.
+  DBSnapshot:
+    Default: ""
+    Description: >-
+      optional, the DB snapshot that is used to restore the DB instance,
+      not used in conjunction with CreateDBName
+    Type: String
+    AllowedPattern: ^[a-zA-Z]*[a-zA-Z0-9\-:]*$
+    ConstraintDescription: >-
+      must begin with a letter and contain only alphanumeric characters, dash,
+      or colon.
+  EnvironmentType:
+    Type: String
+    AllowedValues:
+      - Staging
+      - Production
+  VPCStack:
+    Description: VPC stack to add this cache into
+    Type: String
+    Default: DataVPCStack
+  CloudWatchWarnDiskSpace:
+    Default: 0
+    Description: >-
+      amount of free disk space in bytes that will trigger warning, 0 means no alarm
+    Type: Number
+  CloudWatchWarnCPU:
+    Default: 0
+    Description: percent of CPU that will trigger warning, 0 means no alarm
+    Type: Number
+    MinValue: 0
+    MaxValue: 100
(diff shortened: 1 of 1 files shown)


db/postgresql-instance.yml
```diff
@@ -0,0 +1,219 @@
...
+    Description: The size of the database (Gb)
+    Type: Number
+    MinValue: 20
+    MaxValue: 16384
+    ConstraintDescription: must be between 5 and 16384Gb.
+  DBInstanceClass:
+    Description: The database instance type
+    Type: String
+    Default: db.t2.small
+    ConstraintDescription: must select a valid database instance type.
```
svevang: Removed the list of db types here, since they are always changing. It seems the user can just look up the list of types and plug one in here.

db/postgresql-instance.yml
```diff
@@ -0,0 +1,219 @@
+# db/rds-instance.yml
```
farski: This should match the file name

db/postgresql-instance.yml
```diff
@@ -0,0 +1,219 @@
...
+    ConstraintDescription: >-
+      must begin with a letter and contain only alphanumeric characters, dash,
+      or colon.
+  EnvironmentType:
+    Type: String
+    AllowedValues:
+      - Staging
+      - Production
+  VPCStack:
+    Description: VPC stack to add this cache into
```
farski: Cache?

farski: This parameter name could be a little clearer. Is if a stack name or stack ID?

db/postgresql-instance.yml
```diff
@@ -0,0 +1,219 @@
...
+      AllocatedStorage: !Ref DBAllocatedStorage
+      StorageType: gp2
+      DBInstanceClass: !Ref DBInstanceClass
+      DBSnapshotIdentifier: !If [UseSnapshot, !Ref DBSnapshot, !Ref "AWS::NoValue"]
+      Engine: postgres
+      EngineVersion: !Ref EngineVersion
+      MasterUsername: !If [UseSnapshot, !Ref "AWS::NoValue", !Ref DBUser]
+      MasterUserPassword: !If [UseSnapshot, !Ref "AWS::NoValue", !Ref DBPassword]
+      MultiAZ: false
+      Tags:
```
farski: Would be good to add a `Project` tag, with a parameter to set the value

db/postgresql-instance.yml
```diff
@@ -0,0 +1,219 @@
...
+    - !Not [!Equals [!Ref OpsWarnSnsTopic, ""]]
+    - !Not [!Equals [!Ref CloudWatchWarnDiskSpace, 0]]
+  CreateCloudWatchWarnCPU: !And
+    - !Not [!Equals [!Ref OpsWarnSnsTopic, ""]]
+    - !Not [!Equals [!Ref CloudWatchWarnCPU, 0]]
+  CreateCloudWatchWarnConnections: !And
+    - !Not [!Equals [!Ref OpsWarnSnsTopic, ""]]
+    - !Not [!Equals [!Ref CloudWatchWarnConnections, 0]]
+Resources:
+  RDSSubnetGroup:
```
farski: This resource type [supports tags](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbsubnet-group.html#cfn-rds-dbsubnetgroup-tags); good to always include them when available

db/postgresql-instance.yml
```diff
@@ -0,0 +1,219 @@
...
+      MasterUserPassword: !If [UseSnapshot, !Ref "AWS::NoValue", !Ref DBPassword]
+      MultiAZ: false
+      Tags:
+        - Key: Environment
+          Value: !Ref EnvironmentType
+        - Key: "prx:cloudformation:stack-name"
+          Value: !Ref AWS::StackName
+        - Key: "prx:cloudformation:stack-id"
+          Value: !Ref AWS::StackId
+      DBSubnetGroupName: !Sub Platform-${EnvironmentType}-${AWS::StackName}-db
```
farski: I believe the `DBSUbnetGroup` [return value](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbsubnet-group.html#aws-resource-rds-dbsubnet-group-return-value) is this name, so you should be able to just do a `!Ref RDSSubnetGroup` to get this value, rather than creating the value in both places

svevang:
@farski thanks!  Can you take another look?

db/postgresql-instance.yml
```diff
@@ -0,0 +1,239 @@
...
+            !Sub ${VPCStackName}-Subnet2
+      Tags:
+        - Key: Environment
+          Value: !Ref EnvironmentType
+        - Key: "prx:cloudformation:stack-name"
+          Value: !Ref AWS::StackName
+        - Key: "prx:cloudformation:stack-id"
+          Value: !Ref AWS::StackId
+        - Key: "Project"
+          value: !Ref ProjectTag
```
farski: `Value` not `value`

db/postgresql-instance.yml
```diff
@@ -0,0 +1,239 @@
...
+        - Fn::ImportValue:
+            !Sub ${VPCStackName}-Subnet2
+      Tags:
+        - Key: Environment
+          Value: !Ref EnvironmentType
+        - Key: "prx:cloudformation:stack-name"
+          Value: !Ref AWS::StackName
+        - Key: "prx:cloudformation:stack-id"
+          Value: !Ref AWS::StackId
+        - Key: "Project"
```
farski: Don't need quotes here

db/postgresql-instance.yml
```diff
@@ -0,0 +1,239 @@
...
+      MultiAZ: false
+      Tags:
+        - Key: Environment
+          Value: !Ref EnvironmentType
+        - Key: "prx:cloudformation:stack-name"
+          Value: !Ref AWS::StackName
+        - Key: "prx:cloudformation:stack-id"
+          Value: !Ref AWS::StackId
+        - Key: "Project"
+          value: !Ref ProjectTag
```
farski: `Value` not `value`

db/postgresql-instance.yml
```diff
@@ -0,0 +1,239 @@
...
+      MasterUserPassword: !If [UseSnapshot, !Ref "AWS::NoValue", !Ref DBPassword]
+      MultiAZ: false
+      Tags:
+        - Key: Environment
+          Value: !Ref EnvironmentType
+        - Key: "prx:cloudformation:stack-name"
+          Value: !Ref AWS::StackName
+        - Key: "prx:cloudformation:stack-id"
+          Value: !Ref AWS::StackId
+        - Key: "Project"
```
farski: Don't need quotes here

db/postgresql-instance.yml
farski: Thoughts on this?

svevang:
Thanks, PTAL @farski

Issue #296, opened 2019-02-07
Create Elasticsearch Service template


farski:
For CMS

PR #300, opened 2019-02-15
Add basic Elasticsearch Service template

farski:
Creates a simply Elasticsearch Service deployment template, with limited customization

### DIFF
--- a/etc/elasticsearch.yml
+++ b/etc/elasticsearch.yml
@@ -15,7 +15,7 @@ Parameters:
     Type: String
     AllowedValues:
       - true
-      - false
+      # - false
   EbsVolumeType:
     Type: String
     AllowedValues:
@@ -23,6 +23,10 @@ Parameters:
       - gp2
   EbsVolumeSize:
     Type: Number
+    Description: >
+      EBS volume size is per node. You can calculate the total cluster size for
+      the Amazon ES domain using the following formula –
+      (number of data nodes) * (EBS volume size).
   ElasticsearchVersion:
     Type: String
     AllowedValues:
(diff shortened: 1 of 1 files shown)


farski:
Closes #296

etc/elasticsearch.yml
```diff
@@ -0,0 +1,130 @@
...
+        Statement:
+          - Effect: Allow
+            Principal:
+              AWS: "*"
+            Action: "es:*"
+            Resource: !Sub "arn:aws:es:${AWS::Region}:${AWS::AccountId}:domain/*"
+  # #     AdvancedOptions:
+  # #       String: String
+      EBSOptions:
+        EBSEnabled: !Ref EbsEnabled
```
cavis: Is this really optional though?  In the web console, I can't even select "Instance Storage" - it's greyed out.

etc/elasticsearch.yml
```diff
@@ -0,0 +1,130 @@
...
+    Type: String
+    AllowedValues:
+      - true
+      - false
+  EbsVolumeType:
+    Type: String
+    AllowedValues:
+      - standard
+      - gp2
+  EbsVolumeSize:
```
cavis: Maybe worth noting that this isn't a straightforward number?  The web console has some notes that you should...

> multiply the EBS storage size per node by the number of data nodes in your cluster to calculate the total storage available to your cluster

etc/elasticsearch.yml
farski: The API docs make it seem very optional. For now I kept the parameter, but forced it to always be `true`

PR #301, opened 2019-02-22
DynamoDB table template

cavis:
Creates a dynamodb table.  Optionally:

- Sets an expiration/TTL field
- Creates a role to grant access to an account-id
- Grants access to a specific pre-existing role in that account (maybe will use this later)

### DIFF
--- a/db/dynamodb-table.yml
+++ b/db/dynamodb-table.yml
@@ -1,0 +1,98 @@
+# db/dynamodb-table.yml
+AWSTemplateFormatVersion: "2010-09-09"
+Description: >-
+  Create a simple primary-key-lookup DynamoDB table
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label:
+          default: Table Definition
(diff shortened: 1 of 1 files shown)


db/dynamodb-table.yml
```diff
@@ -0,0 +1,149 @@
...
+    Binary:
+      Symbol: B
+Resources:
+  DynamoDBTable:
+    Type: AWS::DynamoDB::Table
+    Properties:
+      AttributeDefinitions:
+        - AttributeName: !Ref PrimaryKey
+          AttributeType: !FindInMap [DataTypesMap, !Ref PrimaryType, Symbol]
+      # https://github.com/awslabs/cfn-python-lint/issues/509
```
cavis: My local `cfn-lint` picks this up.  But apparently the build passed?

cavis: Also, I went with on-demand pricing for now.  If/when we move enough things to IAB-2.0 that this is expensive, will update to reserved.

PR #302, opened 2019-02-22
analytics-dynamodb-lambda

cavis:
Initialize the analytics-dynamodb-lambda.

- [x] Triggered by the `dovetail-ddb-production` stream
- [x] Writes to dynamodb table `antebytes_production`
- [x] Uses the CrossAccountAccessRole in #301 
- [x] Writes records back to kinesis stream `dovetail-metrics-production-6s`

Can confirm those parameterstore values with:

```
aws ssm get-parameters-by-path --path /prx/stag/analytics-ingest-lambda/
aws ssm get-parameters-by-path --path /prx/prod/analytics-ingest-lambda/
```

### DIFF
--- a/stacks/analytics-ingest-lambda.yml
+++ b/stacks/analytics-ingest-lambda.yml
@@ -18,6 +18,14 @@ Parameters:
     Type: String
   EnvironmentTypeAbbreviation:
     Type: String
+  MetricsKinesisStream:
+    Type: AWS::SSM::Parameter::Value<String>
+  DynamodbKinesisStream:
+    Type: AWS::SSM::Parameter::Value<String>
+  DynamodbTableName:
+    Type: AWS::SSM::Parameter::Value<String>
+  DynamodbAccessRole:
+    Type: AWS::SSM::Parameter::Value<String>
 Mappings:
   VPCSettingsMap:
     test:
@@ -63,6 +71,24 @@ Resources:
                   - "ssm:GetParametersByPath"
                 Resource:
                   - !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/prx/${EnvironmentTypeAbbreviation}/analytics-*"
+        - PolicyName: KinesisWritePolicy
+          PolicyDocument:
+            Version: "2012-10-17"
+            Statement:
+              - Effect: Allow
+                Action:
+                  - "kinesis:DescribeStream"
+                  - "kinesis:PutRecord"
+                  - "kinesis:PutRecords"
+                Resource:
+                  - !Ref MetricsKinesisStream
+        - PolicyName: DynamodbAssumeRolePolicy
+          PolicyDocument:
+            Version: "2012-10-17"
+            Statement:
+              - Effect: Allow
+                Action: "sts:AssumeRole"
+                Resource: !Ref DynamodbAccessRole
       ManagedPolicyArns:
         - "arn:aws:iam::aws:policy/service-role/AWSLambdaKinesisExecutionRole"
         - "arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole"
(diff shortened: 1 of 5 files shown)


stacks/analytics-ingest-lambda.yml
```diff
@@ -18,6 +18,14 @@ Parameters:
     Type: String
```
farski: Since this file has gotten so…robust, I wouldn't hate some comments/docs at the top to give a brief overview of what's all happening

stacks/analytics-ingest-lambda.yml
```diff
@@ -149,6 +175,45 @@ Resources:
...
+    Type: "AWS::Lambda::Function"
+    Properties:
+      Code:
+        S3Bucket: !Ref CodeS3Bucket
+        S3Key: lambda/PRX-analytics-ingest-lambda.zip
+        S3ObjectVersion: !Ref CodeS3ObjectVersion
+      Description: Dovetail analytics to dynamodb
+      Environment:
+        Variables:
+          DYNAMODB: "true"
```
farski: This is fairly opaque/ambiguous variable name

stacks/analytics-ingest-lambda.yml
```diff
@@ -63,6 +71,24 @@ Resources:
                   - "ssm:GetParametersByPath"
                 Resource:
                   - !Sub "arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter/prx/${EnvironmentTypeAbbreviation}/analytics-*"
+        - PolicyName: KinesisWritePolicy
```
farski: I'd consider making a second IAM policy, since the policy needs of the various lambdas are diverging quite a bit

stacks/analytics-ingest-lambda.yml
```diff
@@ -319,3 +384,51 @@ Resources:
       Dimensions:
         - Name: FunctionName
           Value: !Ref AnalyticsRedisLambdaFunction
+  AnalyticsDynamodbErrorAlarm:
+    Type: "AWS::CloudWatch::Alarm"
+    Properties:
+      ActionsEnabled: true
+      AlarmName: !Sub "[AnalyticsLambda][DynamoDB][Errors] ${EnvironmentType} > 0"
```
farski: I've been wondering if `> 0` for lambda errors is too noisy, even for `WARN`, because of the automatic retrying. The first retry happens almost immediately, so `> 1` in 5 minutes would do a decent job of filtering out the transient (single request) issue. Most of the events that trigger these alarms have probably successfully completed before we even get the alarm.

stacks/analytics-ingest-lambda.yml
```diff
@@ -319,3 +384,51 @@ Resources:
...
+      Namespace: AWS/Lambda
+      Period: 300
+      Statistic: Sum
+      Threshold: 0
+      TreatMissingData: notBreaching
+      Dimensions:
+        - Name: FunctionName
+          Value: !Ref AnalyticsDynamodbLambdaFunction
+  AnalyticsDynamodbIteratorAgeAlarm:
+    Type: "AWS::CloudWatch::Alarm"
```
farski: Also thinking maybe a warn on these getting some minutes behind (15?) wouldn't be a bad idea.

stacks/analytics-ingest-lambda.yml
farski: Maybe just a comment to describe what it does for now

stacks/analytics-ingest-lambda.yml
```diff
@@ -149,6 +175,45 @@ Resources:
...
+          Value: !Ref AWS::StackName
+        - Key: "prx:cloudformation:stack-id"
+          Value: !Ref AWS::StackId
+      Timeout: 30
+  # Lambda Triggers
+  # TODO: bigquery/pingbacks/redis triggers manually managed, and tricky to move
+  AnalyticsDynamodbKinesisTrigger:
+    Type: "AWS::Lambda::EventSourceMapping"
+    Properties:
+      BatchSize: 100
```
farski: Any benefit of moving this to SSM so it can be updated without a deploy?

stacks/analytics-ingest-lambda.yml
cavis: Yeah, the way the lambda code decides which of the 4 "modes" to run in is keyed to these kind of odd variables.  I'll update them to some sort of `MODE: "ddb"` / `MODE: "redis"` at some point.  Will open a ticket - requires some coordination between CFN/lambda-code.

cavis: PRX/analytics-ingest-lambda#34

stacks/analytics-ingest-lambda.yml
cavis: It would still require a deploy, right?  But not a code-change to Infrastructure.

PR #304, opened 2019-03-01
WAF sample template

farski:
This was really only to get familiar with the system and have a basic reference

### DIFF
--- a/security/waf-sample.yml
+++ b/security/waf-sample.yml
@@ -85,3 +85,6 @@ Resources:
           PositionalConstraint: STARTS_WITH
           TargetString: mozilla
           TextTransformation: LOWERCASE
+Outputs:
+  WebAclId:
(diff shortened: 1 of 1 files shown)


cavis:
So it seems like we'll have to (a) maintain at least 2 of these (for CF vs ALBs), and (b) pretty much all DIY on the rules.  The examples aren't all that helpful ... like "<SQL injection match condition here>".

Trying to imagine what sort of workflow we'd use to detect/block malicious traffic using this.  All I can think of so far is "check the server logs / see offending IP(s) / block".  Which I guess is better than nothing.

farski:
This isn't a WIP, or at least I have no more work I'm planning on doing for it. If folks are done reviewing it as a sample/example we can merge or just close it.

cavis:
(Though there does seem to be an unrelated linter failure)

PR #305, opened 2019-03-06
Tune alarms

cavis:


cavis:
@farski 👍to the CI/CD lambda thresholds.

stacks/analytics-ingest-lambda.yml
```diff
@@ -213,10 +213,10 @@ Resources:
       AlarmDescription:
         Too few invocations
       ComparisonOperator: LessThanThreshold
-      EvaluationPeriods: 1
+      EvaluationPeriods: 2
       MetricName: Invocations
       Namespace: AWS/Lambda
-      Period: 300
+      Period: 600
```
farski: This will already make it twice as hard to hit (< 1500 in 10m vs <1500 in 5m), and it will need to happen twice in a row, up from once. Just want to confirm we're looking for that big of a swing

cavis: Ah, you're right.  I meant to change `EvaluationPeriods -> 3`, as I'm more concerned if it keeps happening.

PR #306, opened 2019-03-06
Dovetail CDN deploy pipeline

farski:


farski:
@cavis No time pressure on this, but curious to get your initial reaction. This isn't super practical as-is, since you have to redeploy the pipeline stack to update the functions' ARNs. Those should just get queried from the API. Pretend that's happening when you evaluate.

farski:
@cavis and I have reviewed this concept. He's going to do a more thorough review of the code, and then we have to get a couple other things done before we can start testing. We may need to add some additional notifications before this is ready for primetime.

Once we're ready, we'll do a test of the pipeline against the staging CloudFront, and verify everything is interacting as expected.

etc/dovetail-edge-deploy.yml
```diff
@@ -0,0 +1,865 @@
...
+    Type: String
+  ProductionStitchLambdaFunctionArn:
+    Description: >
+      e.g., arn:aws:lambda:<region>:<account>:function:<name>
+    Type: String
+  ProductionBytesLambdaFunctionArn:
+    Description: >
+      e.g., arn:aws:lambda:<region>:<account>:function:<name>
+    Type: String
+  CertificateArn:
```
cavis: Do dovetail staging/prod share a cert?

etc/dovetail-edge-deploy.yml
```diff
@@ -0,0 +1,865 @@
...
+    Type: AWS::CodePipeline::Pipeline
+    Properties:
+      ArtifactStore:
+        Location: !Ref PipelineArtifactStore
+        Type: S3
+      RoleArn: !GetAtt PipelineIamRole.Arn
+      Stages:
+        - Name: Source
+          Actions:
+            - Name: InfrastructureRepo
```
cavis: The Infrastructure repo is an input to this?  That seems odd - isn't this edge-deploy stack manual?

etc/dovetail-edge-deploy.yml
```diff
@@ -0,0 +1,866 @@
...
+                Category: Invoke
+                Owner: AWS
+                Provider: Lambda
+                Version: "1"
+              Configuration:
+                FunctionName: !Ref DiscoverEdgeLambdaVersionsFunction
+              OutputArtifacts:
+                - Name: LambdaVersionsArtifact
+              RunOrder: 2
+        # - Name: Staging
```
cavis: Will need to uncomment the staging changeset bit.

etc/dovetail-edge-deploy.yml
```diff
@@ -0,0 +1,866 @@
...
+                  jobId=job['id'],
+                  failureDetails={'message': message, 'type': 'JobFailed'})
+
+
+          # Tries to publish new versions of each Lambda function, from their $LATEST
+          # code and configuration. New versions are only created if a any changes have
+          # been made since the most recent version, so it is safe to call this
+          # multiple times
+          def publish_versions():
+              lamb.publish_version(FunctionName=os.environ['STAGING_BYTES_LAMBDA_FUNCTION_ARN'])
```
cavis: Do these throw an error, if the code hasn't changed?

etc/dovetail-edge-deploy.yml
```diff
@@ -0,0 +1,866 @@
...
+              )
+
+
+              production_bytes = lamb.list_versions_by_function(
+                  FunctionName=os.environ['PRODUCTION_BYTES_LAMBDA_FUNCTION_ARN'],
+                  MaxItems=999
+              )
+
+              return {
+                'staging_bytes': staging_bytes['Versions'][-1]['Version'],
```
cavis: So the returned order is guaranteed to be oldest to newest?

etc/dovetail-edge-deploy.yml
```diff
@@ -0,0 +1,865 @@
...
+        default: Origin path
+      ProductionDistributionDomain:
+        default: Distribution domain
+      BytesLambdaFunctionArn:
+        default: Unqualified Bytes Lambda function ARN
+      StitchLambdaFunctionArn:
+        default: Unqualified Stitch Lambda function ARN
+Parameters:
+  StagingOriginBucket:
+    Default: prx-dovetail.s3.amazonaws.com
```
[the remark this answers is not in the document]
farski: Rather than passing them in here as parameter overrides, we could make a template config JSON file that lives somewhere for the more permanent values. This seems like a good place to start though without having to shuffle files around in S3, and we can switch it up later if this feels brittle.

etc/dovetail-edge-deploy.yml
```diff
@@ -0,0 +1,865 @@
...
+    Type: String
+  InfrastructureNotificationsStackName:
+    Default: infrastructure-notifications
+    Type: String
+  GitHubToken:
+    Description: >
+      A GitHub access token with permissions to update statuses, request
+      content, etc
+    Type: String
+  GitHubWebhookSecret:
```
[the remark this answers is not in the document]
farski: For the time being I've turned off the GitHub webhooks, so none of the triggers on this pipeline should actually automatically trigger. This was mostly a carry over from copying and pasting the CD template

etc/dovetail-edge-deploy.yml
farski: I think they used to, and in April we added a cert to dovetail-cdn.yml that it manages on its own, so this should be removed.

etc/dovetail-edge-deploy.yml
farski: It needs to be an input so we can get the YAML file to pass to Cfn to deploy

etc/dovetail-edge-deploy.yml
farski: No, it's a no op

etc/dovetail-edge-deploy.yml
farski: I don't see that explicitly stated anywhere, but I've never seen it do anything else

etc/dovetail-edge-deploy.yml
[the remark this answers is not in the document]
cavis: +1

etc/dovetail-edge-deploy.yml
cavis: Might be a reason to always show changesets in slack (or at least "going from version X -> version Y") before you click the prod deploy button.  Just in case something wacky happens.

cavis: But yeah, can tackle that later, if this works.

PR #311, opened 2019-03-28
Infrastructure deploy scripts

farski:
Starting to work on some deploy scripts for the larger aspects of Infrastructure.

Right now these really only work for ***updating*** stacks, because they don't require (or even allow) for all of the stack parameters to be defined. Mostly looking for feedback on the general approach.

The main benefit to this is that it makes it much easier to update the Lambda functions that are used part of ci/cd/etc.

FWIW this is based off off a deploy script I use for some other projects, which are primarily serve


farski:
@cavis would like to have you eyes on this; no rush

farski:
This is better solved with SAM, so I'm closing it.

PR #314, opened 2019-04-05
Add certificate to feeder-cdn

farski:


cdn/feeder-cdn.yml
```diff
@@ -29,6 +27,20 @@ Parameters:
     Type: String
     Description: eg. arn:aws:acm:<region>:<account>:certificate/<guid>
 Resources:
+  Certificate:
+    Type: "AWS::CertificateManager::Certificate"
+    Properties:
+      DomainName: !Ref DistributionDomain
```
cavis: I'm noting that currently both `https://feeder.prx.org` and `https://feeder.prx.tech` are valid.  Not sure we're using the latter anymore, but we have used it with dovetail in the past.  So this would be a bit of a regression.

May want to throw both in there for prod, if it's easy.

cavis: Oh wait, this isn't actually feeder... `f.prxu.org`!  Approved!

PR #319, opened 2019-04-24
Pixel tracker

cavis:
Initial deployment for a pixel tracker API gateway [pixel.prx.org](https://github.com/PRX/pixel.prx.org).

This goes the same route as proxy.prx.org, and just sends _all_ HTTP methods and paths on to the lambda.  May dial that back at some point, to reduce load.  But this seemed like an easy place to start.

### DIFF
--- a/stacks/serverless/pixel.prx.org.yml
+++ b/stacks/serverless/pixel.prx.org.yml
@@ -1,0 +1,231 @@
+# stacks/serverless/pixel.prx.org.yml
+AWSTemplateFormatVersion: "2010-09-09"
+Description: API Gateway pixel tracker
+Conditions:
+  CreateStagingResources: !Equals [!Ref EnvironmentType, Staging]
+  CreateProductionResources: !Equals [!Ref EnvironmentType, Production]
+Parameters:
+  OpsWarnMessagesSnsTopicArn:
+    Type: String
+  OpsErrorMessagesSnsTopicArn:
+    Type: String
+  CodeS3Bucket:
+    Type: String
+  CodeS3ObjectVersion:
+    Type: String
+  EnvironmentType:
+    Type: String
+  EnvironmentTypeAbbreviation:
+    Type: String
+Mappings:
+  EnvironmentTypeMap:
+    Testing:
+      DomainName: "pixel.test.prx.tech"
+      HostedZoneName: "prx.tech."
+      ValidationDomain: "prx.tech"
+    Staging:
+      DomainName: "pixel.staging.prx.tech"
+      HostedZoneName: "prx.tech."
+      ValidationDomain: "prx.tech"
+    Production:
+      DomainName: "pixel.prx.org"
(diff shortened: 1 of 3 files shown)


stacks/serverless/pixel.prx.org.yml
```diff
@@ -0,0 +1,228 @@
...
+      CertificateArn: !Ref Certificate
+      DomainName: !FindInMap [EnvironmentTypeMap, !Ref EnvironmentType, DomainName]
+  PixelRestApiBasePathMapping:
+    Type: "AWS::ApiGateway::BasePathMapping"
+    Properties:
+      DomainName: !Ref PixelRestApiDomainName
+      RestApiId: !Ref PixelRestApi
+      Stage: prod
+  PixelRestApiDomainNameRecordSetGroup:
+    Type: "AWS::Route53::RecordSetGroup"
```
cavis: Copied this out of the proxy stack ... so I guess we're creating DNS records within the CD stack now?  Or maybe that's just the way you have to do things with the weird API-Gateway cloudfront/dns setup.

stacks/serverless/pixel.prx.org.yml
```diff
@@ -0,0 +1,228 @@
...
+      DomainName: "pixel.staging.prx.tech"
+      HostedZoneName: "prx.tech."
+      ValidationDomain: "prx.tech"
+    Production:
+      DomainName: "pixel.prx.org"
+      HostedZoneName: "prx.org."
+      ValidationDomain: "prx.org"
+Resources:
+  Certificate:
+    Type: "AWS::CertificateManager::Certificate"
```
farski: I think we should do this as DNS validation, rather than email. [Ex](https://github.com/PRX/Infrastructure/blob/master/etc/wildcard-certificate.yml#L28)

stacks/serverless/pixel.prx.org.yml
farski: We have done this in a couple cases. I can take it or leave it, the `prx.tech` hosted zone is not templated, and it would be a massive pain to convert it, so whether we add the entries manually or in these stacks, not sure it matters much. We'll hit some pain if we ever try to deploy multiple instances, though

stacks/serverless/pixel.prx.org.yml
cavis: Ah, right.  Didn't know that needed a setting.

stacks/serverless/pixel.prx.org.yml
```diff
@@ -0,0 +1,229 @@
...
+    Properties:
+      DomainName: !FindInMap [EnvironmentTypeMap, !Ref EnvironmentType, DomainName]
+      DomainValidationOptions:
```
farski: And actually we don't need this if we're doing DNS

cavis:
Thanks - i'll take care of the whole first-deployment coordination thing.

PR #320, opened 2019-04-26
Styleguide stack

pstjean:
Based on how Metrics works


stacks/styleguide.prx.org.yml
```diff
@@ -0,0 +1,208 @@
...
+    Type: String
+  # Styleguide ENV ################################################################
+  StyleguideSecretsVersion:
+    Type: String
+Resources:
+  StyleguideLogGroup:
+    Type: "AWS::Logs::LogGroup"
+    Properties:
+      RetentionInDays: 14
+  StyleguideALBTargetGroup:
```
pstjean: I'm not sure whether Styleguide needs any load balancing at all, but I've retained it from Metrics.

stacks/styleguide.prx.org.yml
```diff
@@ -0,0 +1,208 @@
...
+            - HostPort: 0
+              ContainerPort: 80
+  StyleguideService:
+    Type: "AWS::ECS::Service"
+    Properties:
+      Cluster: !Ref ECSCluster
+      DeploymentConfiguration:
+        MaximumPercent: 200
+        MinimumHealthyPercent: 100
+      DesiredCount: 1
```
pstjean: I've lowered this to a single container.

stacks/styleguide.prx.org.yml
```diff
@@ -0,0 +1,208 @@
...
+        - Cpu: !Ref ContainerCpu
+          Environment:
+            - Name: APP_NAME
+              Value: "styleguide"
+            - Name: APP_ENV
+              Value: !Ref EnvironmentTypeAbbreviation
+            - Name: AWS_SECRETS_BASE
+              Value: !Ref SecretsBase
+            - Name: AWS_SECRETS_VERSION
+              Value: !Ref StyleguideSecretsVersion
```
pstjean: I'm fuzzy on how the secrets are set up, is this part of the process of creating the initial configuration in S3 described here? https://github.com/PRX/Infrastructure/tree/master/cd#staging-deploy

stacks/root.yml
```diff
@@ -900,6 +907,48 @@ Resources:
           Value: !Ref AWS::StackId
       TemplateURL: !Join ["", ["http://s3", !If [IsUsEast1, "", "-"], !If [IsUsEast1, "", !Ref "AWS::Region"], ".amazonaws.com/", "Fn::ImportValue": !Sub "${InfrastructureStorageStackName}-InfrastructureSourceBucket", "/", !Ref InfrastructureGitCommit, "/stacks/proxy.prx.org.yml"]]
       TimeoutInMinutes: 5
+  StyleguideStack:
+    Type: "AWS::CloudFormation::Stack"
+    Condition: "IsStaging"
```
pstjean: I happened to see this in the PR to add Metrics, and it seemed like a good idea, so I swiped it.

stacks/styleguide.prx.org.yml
cavis: That's the S3 object version set automagically by CD - coming out of the `s3://prx-infrastructure-us-east-1-config/template-config-staging.zip`.  And there's some sort of lambda trigger on the S3 bucket that will update the zip whenever you change the secrets file in S3.

stacks/root.yml
cavis: +1.  Staging-only should work fine for styleguide.  Save some ECS space.

stacks/container-apps/root.yml
```diff
@@ -340,6 +347,50 @@ Resources:
...
+          Fn::ImportValue:
+            !Sub "${InfrastructureNotificationsStackName}-OpsDebugMessagesSnsTopicArn"
+        OpsErrorMessagesSnsTopicArn:
+          Fn::ImportValue:
+            !Sub "${InfrastructureNotificationsStackName}-OpsErrorMessagesSnsTopicArn"
+        EnvironmentType: !Ref EnvironmentType
+        EnvironmentTypeAbbreviation: !Ref EnvironmentTypeAbbreviation
+        EcrRegion: !Ref EcrRegion
+        SecretsBase: !Ref SecretsBase
+        ContainerMemory: !Ref ContainerMemory
```
cavis: Worth noting these are the "Backend" memory/cpu reservation numbers in the root stack.  So may be a bit overpowered for a frontend app.  But it's staging, and I don't think worth changing these.

May want to set `DesiredWebs` to 1?  But also not important.  Staging apps are typically underpowered anyways.

pstjean: `DesiredWebs` makes good sense, I had made that change in the old version of the config.

PR #322, opened 2019-05-01
Dovetail Redis cluster template

farski:


farski:
I'm closing this. If we find that we need a Redis template we can start over with three more years of Cfn knowledge.

PR #326, opened 2019-06-04
Add in networks app

kookster:


stacks/container-apps/root.yml
```diff
@@ -172,6 +179,52 @@ Resources:
...
+        ContainerCpu: !Ref ContainerCpu
+        PlatformALBListenerPriorityPrefix: !Ref NetworksPlatformALBListenerPriorityPrefix
+        EcrImageTag: !Ref NetworksEcrImageTag
+        SecretsVersion: !Ref NetworksSecretsVersion
+        AppName: "networks"
+        CreateWorker: "false"
+        DesiredWebsProduction: "1"
+        ErrorAlarmThreshold: "5"
+        ContainerPort: "3000"
+        HealthCheckPath: "/sessions/new"
```
cavis: May want to do a middleware health check, at some point.  But this works!

kookster: just getting it working for now, but yes, that would be better, like exchange.

PR #328, opened 2019-06-12
Add sphinx service and Network Load Balancer

kookster:
I think I have everything I need? not entirely sure, and the `Protocol: TLS` isn't validating locally in cfn-lint even though the docs say that is supported for https endpoints.

Anyway, pls have a look.

### DIFF
--- a/stacks/container-apps/root.yml
+++ b/stacks/container-apps/root.yml
@@ -17,6 +17,14 @@ Parameters:
   # VPC ########################################################################
   VPC:
     Type: "AWS::EC2::VPC::Id"
+  VPCSubnet1:
+    Type: "AWS::EC2::Subnet::Id"
+  VPCSubnet2:
+    Type: "AWS::EC2::Subnet::Id"
+  VPCSubnet3:
+    Type: "AWS::EC2::Subnet::Id"
+  VPCCertificateArn:
+    Type: String
   # Load Balancer ##############################################################
   PlatformALBDNSName:
     Type: String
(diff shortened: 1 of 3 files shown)


stacks/container-apps/root.yml
```diff
@@ -225,6 +233,52 @@ Resources:
...
+            !Sub "${InfrastructureNotificationsStackName}-OpsWarnMessagesSnsTopicArn"
+        OpsErrorMessagesSnsTopicArn:
+          Fn::ImportValue:
+            !Sub "${InfrastructureNotificationsStackName}-OpsErrorMessagesSnsTopicArn"
+        EnvironmentType: !Ref EnvironmentType
+        EnvironmentTypeAbbreviation: !Ref EnvironmentTypeAbbreviation
+        SecretsBase: !Ref SecretsBase
+        EcrRegion: !Ref EcrRegion
+        EcrImageTag: !Ref NetworksEcrImageTag
+        DesiredWebs: "1"
```
cavis: Could probably just hardcode some of this stuff in the sub-stack, since the Sphinx stack will only ever be used for this 1 purpose.  But this is also fine.

stacks/container-apps/sphinx-nlb.yml
```diff
@@ -0,0 +1,300 @@
...
+      TreatMissingData: notBreaching
+      Dimensions:
+        - Name: LoadBalancer
+          Value: !GetAtt NLB.LoadBalancerFullName
+  # ECS Service - Web
+  WebLogGroup:
+    Type: "AWS::Logs::LogGroup"
+    Properties:
+      RetentionInDays: 14
+  WebTaskDefinition:
```
cavis: So this task won't actually be serving web traffic, right?  (It's not going to be in the web ALB)

stacks/container-apps/sphinx-nlb.yml
```diff
@@ -0,0 +1,300 @@
+# stacks/web-alb-application.yml
```
cavis: Should probably change this

stacks/container-apps/sphinx-nlb.yml
```diff
@@ -0,0 +1,300 @@
...
+  NLBSphinxListener:
+    Type: "AWS::ElasticLoadBalancingV2::Listener"
+    Properties:
+      LoadBalancerArn: !Ref NLB
+      Port: 9312
+      Protocol: TCP
+      DefaultActions:
+        - Type: forward
+          TargetGroupArn: !Ref NLBTargetGroup
+  NLBHTTPListener:
```
farski: If Exchange is the only thing hitting Sphinx, it may not need both the HTTP and HTTPS listeners

stacks/container-apps/sphinx-nlb.yml
farski: Should be `stacks/container-apps/sphinx-nlb.yml`

kookster: oops

kookster: yes, changing it

stacks/container-apps/sphinx-nlb.yml
kookster: It is hit by networks, not exchange, but yes, only by networks.
I was thinking maybe we only do http but in general having https available seemed like a good thing?

stacks/container-apps/sphinx-nlb.yml
kookster: It is serving web traffic, for the requests to process the delta index.

stacks/container-apps/sphinx-nlb.yml
kookster: fixed

kookster: fixed

PR #333, opened 2019-07-08
adds Grove back

sandikbarr:


stacks/container-apps/root.yml
```diff
@@ -573,47 +573,47 @@ Resources:
...
+          Fn::ImportValue:
+            !Sub "${InfrastructureNotificationsStackName}-OpsDebugMessagesSnsTopicArn"
+        OpsErrorMessagesSnsTopicArn:
+          Fn::ImportValue:
+            !Sub "${InfrastructureNotificationsStackName}-OpsErrorMessagesSnsTopicArn"
+        EnvironmentType: !Ref EnvironmentType
+        EnvironmentTypeAbbreviation: !Ref EnvironmentTypeAbbreviation
+        EcrRegion: !Ref EcrRegion
+        SecretsBase: !Ref SecretsBase
+        ContainerMemory: !Ref ContainerMemory
```
cavis: Note to self: at some point, we should start using less memory for frontend apps.  Maybe passing through `FrontendCpu` and `FrontendMemory` from the `stacks/root.yml`, and using those here.  But that's a separate issue.

stacks/container-apps/root.yml
```diff
@@ -573,47 +573,47 @@ Resources:
...
+        ContainerMemoryReservation: !Ref ContainerMemoryReservation
+        ContainerCpu: !Ref ContainerCpu
+        PlatformALBListenerPriorityPrefix: !Ref GrovePlatformALBListenerPriorityPrefix
+        EcrImageTag: !Ref GroveEcrImageTag
+        SecretsVersion: !Ref GroveSecretsVersion
+        AppName: "grove"
+        CreateWorker: "false"
+        ContainerPort: "4204"
+        HealthCheckPath: "/api/v1"
```
cavis: Does this path really exist for this app?  May just want `"/"` (or anything that returns a 200).

stacks/container-apps/root.yml
cavis: Opened #334

stacks/container-apps/root.yml
sandikbarr: nope, not sure what jingle is doing with it. _so many details_

PR #338, opened 2019-08-16
HOLD Allow detached CDN certificates

farski:
🚧 HOLD for now.

**These changes should (must) all be backwards compatible with the previous version of the template.**

Adds a `CertificateAttachment` parameter to the stack, which can be set to either `Attached` or `Detached`.

When set to `Attached` the template should behave as it did previously.

When set to `Detached`, the CloudFront distribution will use the default `*.cloudfront.net` certificate, rather than using the custom ACM certificate created by the stack, which will eliminate the dependency between the distribution and the certificate. This will allow the distribution to get created immediately along with the certificate, rather than waiting for the certificate to validate. 

Because CF aliases can no longer be added to distributions without a matching validated certificate, when `Detached` the aliases are not included.

The CloudFormation stack will not completely deploy until the certificate has been validated, ~but I think that certificate resources have an unlimited timeout, and the stack creation itself also has no timeout, so that should be fine even if it takes hours to days.~ The ACM certificate resource timeout after 12 hours if they're not validated.

In practice, this will allow us to provide both the validation DNS as well as the CF DNS records to the end user in one step. Once the certificate is validated, the stack can be switched to `Attached`, which will add the aliases to the distribution, and associated the custom certificate with the distribution.

The creation of the certificate resource is still solely dependent on whether the ARN of a preexisting certificate is provided in the `AcmCertificateArn` parameter.

### DIFF
--- a/cdn/single-origin.yml
+++ b/cdn/single-origin.yml
@@ -36,10 +36,16 @@ Conditions:
   HasCloudFrontWebACLId: !Not [!Equals [!Ref CloudFrontWebACLId, ""]]
   HasAcmCertificateArn: !Not [!Equals [!Ref AcmCertificateArn, ""]]
   HasNoAcmCertificateArn: !Equals [!Ref AcmCertificateArn, ""]
+  HasDetachedCertificate: !Equals [!Ref CertificateAttachment, "Detached"]
   HasOriginPath: !Not [!Equals [!Ref OriginPath, ""]]
 Metadata:
   AWS::CloudFormation::Interface:
     ParameterGroups:
+      - Label:
+          default: HTTPS Certificate
+        Parameters:
+          - CertificateAttachment
+          - AcmCertificateArn
       - Label:
           default: Distribution
         Parameters:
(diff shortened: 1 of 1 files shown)


cdn/single-origin.yml
```diff
@@ -113,58 +115,74 @@ Metadata:
...
+    AllowedPattern: ^.+\.[a-zA-Z]+$
+    ConstraintDescription: >
+      must be a hostname or of the format mybucket.s3.amazonaws.com
     Description: >
       The DNS name of the Amazon Simple Storage Service (S3) bucket or the HTTP
       server from which you want CloudFront to get objects for this origin.
+      (e.g. "example.com" or "mybucket.s3.amazonaws.com")
   OriginPath:
     Type: String
+    AllowedPattern: ^$|^\/..+(?<!\/)$
```
farski: Does this look right?

cavis: Hrmm... in a normal language I would have written that as `/^\/.+[^\/]$` (starts with `/`, followed by 0 or more of anything, ending with not-`/`).  Do those `AllowedPattern`s have a funny format or something?

farski: I'm just bad at regular expressions. Your way seems better. I think I had the extra dot in there to make sure it contained a dot.

cavis:
Moving this back to "draft", since it may never merge.

farski:
The only real benefit of this was to be able to get the CloudFront distribution's domain name before the cert's validation records existed. The 12 hour timeout on creating the cert mean there still needs to close coordination to create the stack.

The same thing can be done by launching the stack without any CNAMEs, and updating the stack whenever the cert can be made. This change was confusing and overly complicated and didn't even solve the problem.

wontfix

PR #340, opened 2019-08-22
Exchange worker autoscaling

cavis:
For PRX/exchange.prx.org#508

- Adds fatal alarms for `Prx3ProductionDeliveryUpdate` and `Prx3ProductionSayWhen`
    _(We actually already had some alarms, but they were ops-error, and the thresholds may have been a bit too lenient.  Will delete them after this deploys)_
- Adds autoscaling to the ECS Worker tasks, but not the say_when task, which I _think_ per PRX/exchange.prx.org#446 has to be a singleton.  (Running more than one would result in duplicate tasks getting scheduled)

The scaling rules I went with, using Prx3ProductionDeliveryUpdate ApproximateNumberOfMessagesVisible:

- If between 0 and 1 once in 1 minute, scale in by 2
- If between 1 and 10 once in 1 minute, scale in by 1
- If between 20 and 100 once in 1 minute, scale out by 2 (cooldown 2 minutes)
- If between 100 and 400 once in 1 minute, scale out by 4 (cooldown 2 minutes)
- If above 400 once in 1 minute, scale out by 6 (cooldown 2 minutes)

Will have to watch these, and see if they behave okay.  I wanted the 1 minute alarm so these would scale quickly.  May need a longer cooldown though.

### DIFF
--- a/stacks/container-apps/web-application.yml
+++ b/stacks/container-apps/web-application.yml
@@ -377,6 +377,9 @@ Outputs:
   WorkerTaskDefinitionName:
     Description: Name of the worker task definition
     Value: !Sub ${AppName}-worker
+  WorkerServiceName:
+    Description: Name of the worker service
+    Value: !If [CreateWorkerResources, !GetAtt WorkerService.Name, ""]
   HostedZoneDNSName:
     Description: Convenience domain name for the ALB in a hosted zone
     Value: !Sub |
--- a/stacks/container-apps/exchange-alarms.yml
+++ b/stacks/container-apps/exchange-alarms.yml
@@ -1,0 +1,172 @@
+# stacks/container-apps/exchange-alarms.yml
+AWSTemplateFormatVersion: "2010-09-09"
+Description: Exchange alarms
+Conditions:
+  CreateProductionResources: !Equals [!Ref EnvironmentType, Production]
+Parameters:
+  OpsWarnMessagesSnsTopicArn:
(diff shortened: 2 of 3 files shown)


kookster:
I could be wrong, but I thought that was row locking on say_when so multiple could run...

yeah:
https://github.com/kookster/say_when/blob/v0.3.0/lib/say_when/storage/active_record/job.rb#L28-L46

Looks like that isn't quite right to me though - so probably better to just run one.

cavis:
IIRC, the first time we turned on the Exchange ECS workers, we got duplicates of every say_when job.  Should probably test and fix that though, to (A) simplify the Infrastructure code and (B) get _all_ the workers under autoscaling, in case SQS-disconnects are a thing.

kookster:
I am pretty sure I see the problem in say_when, where it does lock a record, but doesn't correctly handle when a record is locked, if that makes any sense.

stacks/container-apps/web-application.yml
```diff
@@ -377,6 +377,9 @@ Outputs:
   WorkerTaskDefinitionName:
     Description: Name of the worker task definition
     Value: !Sub ${AppName}-worker
+  WorkerServiceName:
+    Description: Name of the worker service
+    Value: !If [CreateWorkerResources, !GetAtt WorkerService.Name, ""]
```
kookster: useful addition.

Issue #354, opened 2019-10-15
Dovetail counts lambda alarms


cavis:
Check that we have at least some timeout/error alarms.  If not a kinesis alarm as well.

PR #355, opened 2019-10-15
Dovetail Counts Lambda alarms

cavis:
For #354.

### DIFF
--- a/stacks/serverless/dovetail-counts-lambda.yml
+++ b/stacks/serverless/dovetail-counts-lambda.yml
@@ -1,12 +1,15 @@
 # stacks/serverless/dovetail-counts-lambda.yml
 AWSTemplateFormatVersion: "2010-09-09"
 Description: Dovetail counts lambda function
-# Conditions:
-#   HasRedisBackupUrl: !Not [!Equals [!Ref RedisBackupUrl, ""]]
+Conditions:
+  IsProduction: !Equals [!Ref EnvironmentType, Production]
+  # HasRedisBackupUrl: !Not [!Equals [!Ref RedisBackupUrl, ""]]
 Parameters:
   OpsWarnMessagesSnsTopicArn:
     Type: String
   OpsErrorMessagesSnsTopicArn:
+    Type: String
+  OpsFatalMessagesSnsTopicArn:
     Type: String
   CodeS3Bucket:
     Type: String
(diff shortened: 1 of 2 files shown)


stacks/serverless/dovetail-counts-lambda.yml
```diff
@@ -125,3 +128,115 @@ Resources:
       EventSourceArn: !Ref KinesisIn
       FunctionName: !Ref DovetailCountsLambdaFunction
       StartingPosition: "LATEST"
+  # log metrics
+  DovetailCountsLambdaWarnsMetric:
+    Type: "AWS::Logs::MetricFilter"
+    Properties:
+      FilterPattern: '{ $._logLevel = "warn" }'
+      LogGroupName: !Sub "/aws/lambda/${DovetailCountsLambdaFunction}"
```
cavis: Is this going to work?

I also thought about specifically defining the `AWS::Logs::LogGroup` for the lambda in here.  (So I could also add an expiration).  But if I recall, you can't add existing log groups to CFN without some monkeying around.  Pausing the lambda, deleting the log group, etc.

farski: Should work since the log group already exists. I'm not sure what would happen if this were a virgin stack and you tried to make a metric filter for a log group that didn't exist

stacks/serverless/dovetail-counts-lambda.yml
```diff
@@ -125,3 +128,115 @@ Resources:
...
+        - !If [IsProduction, !Ref OpsErrorMessagesSnsTopicArn, !Ref OpsWarnMessagesSnsTopicArn]
+      AlarmDescription:
+        Too many dovetail counts lambda errors
+      ComparisonOperator: GreaterThanThreshold
+      EvaluationPeriods: 1
+      MetricName: Errors
+      Namespace: AWS/Lambda
+      Period: 60
+      Statistic: Sum
+      Threshold: 0
```
farski: Are we relying on native Lambda retries for anything with this? If so, the `0` threshold is probably too low

stacks/serverless/dovetail-counts-lambda.yml
farski: It should be fine since the log group already exists. I'm not sure what happens if you try to make a metric filter for a nonexistent log group.

PR #356, opened 2019-10-17
Configure NLB to have 2 target groups, and the ecs service to use both

kookster:
I think this will correctly setup 2 target groups on one NLB, each with their own port, and then map an ECS service to connect to both.

### DIFF
--- a/stacks/container-apps/sphinx-nlb.yml
+++ b/stacks/container-apps/sphinx-nlb.yml
@@ -68,161 +68,187 @@ Parameters:
     Type: String
     Default: "/sessions/new"
 Resources:
-  # NLB:
-  #   Type: "AWS::ElasticLoadBalancingV2::LoadBalancer"
-  #   Properties:
-  #     IpAddressType: ipv4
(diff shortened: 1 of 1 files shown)


stacks/container-apps/sphinx-nlb.yml
```diff
@@ -68,161 +68,187 @@ Parameters:
...
+      Certificates:
+        - CertificateArn: !Ref VPCCertificateArn
+      LoadBalancerArn: !Ref NLB
+      Port: 443
+      Protocol: TLS
+      SslPolicy: "ELBSecurityPolicy-2016-08"
+      DefaultActions:
+        - Type: forward
+          TargetGroupArn: !Ref NLBTargetGroup
+  NLBTargetGroup:
```
farski: Maybe this could be called `NLBWebTargetGroup` or something more distinct

kookster: yeah, and I think I have a mistake in using these with the listeners...

PR #358, opened 2019-10-25
Dovetail elixir

cavis:
Setup [dovetail-elixir](https://github.com/PRX/dovetail-elixir) app under the Dovetail ALB.

- [x] Add service/task-def/target-group for dovetail-elixir
- [x] Add listener rules to the dovetail ALB

Will need to merge PRX/dovetail-elixir#7 first.

### DIFF
--- a/stacks/dovetail.prx.org.yml
+++ b/stacks/dovetail.prx.org.yml
@@ -51,6 +51,11 @@ Parameters:
     Type: String
   DovetailSecretsVersion:
     Type: String
+  # Dovetail Elixir ############################################################
+  DovetailEcrImageTag:
+    Type: String
+  DovetailSecretsVersion:
+    Type: String
 Mappings:
   EnvironmentTypeMap:
     Testing:
@@ -289,7 +294,7 @@ Resources:
       DeploymentConfiguration:
         MaximumPercent: 200
         MinimumHealthyPercent: 50
-      DesiredCount: !GetAtt DovetailServiceDesiredCountRolloverResource.DesiredCount
+      DesiredCount: 10
       LoadBalancers:
         - ContainerName: "dovetail-express"
           ContainerPort: 8080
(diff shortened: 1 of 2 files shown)


stacks/dovetail.prx.org.yml
```diff
@@ -124,6 +129,30 @@ Resources:
...
+  DovetailALBHTTPListenerRule:
+    Type: "AWS::ElasticLoadBalancingV2::ListenerRule"
+    Properties:
+      Actions:
+        - TargetGroupArn: !Ref DovetailApplicationALBTargetGroup
+          Type: forward
+      Conditions:
+        - Field: host-header
+          Values:
+            - dovetail.*
```
cavis: IIRC, you get charged based on how many listener rules a request goes through.  So this sets up any `dovetail.prxu.org` requests as the first rule.  Then `dovetail3.*` requests.  Then the default (also the old Dovetail node app).

stacks/dovetail.prx.org.yml
```diff
@@ -289,7 +354,7 @@ Resources:
       DeploymentConfiguration:
         MaximumPercent: 200
         MinimumHealthyPercent: 50
-      DesiredCount: !GetAtt DovetailServiceDesiredCountRolloverResource.DesiredCount
+      DesiredCount: 10
```
cavis: Per the exchange-worker autoscaling PR, you no longer have to worry about this blowing away any autoscaling changes to the desired count.

stacks/dovetail.prx.org.yml
```diff
@@ -350,72 +415,98 @@ Resources:
...
-      ServiceToken: !GetAtt DovetailServiceDesiredCountRolloverLambdaFunction.Arn
+      RetentionInDays: 14
+  DovetailElixirTaskDefinition:
+    Type: "AWS::ECS::TaskDefinition"
+    Properties:
+      ContainerDefinitions:
+        - Cpu: !Ref ContainerCpu
+          Environment:
+            - Name: APP_NAME
+              Value: "dovetail-elixir"
```
cavis: I _think_ you can set the app-name as whatever you want in aws-secrets ... even including dashes.

stacks/dovetail.prx.org.yml
farski: Want to remove that rollover custom resource in this same PR?

farski: JK you did

PR #360, opened 2019-10-29
Reprioritize ALB listener rules

### LINKED ISSUE #359: Reorder ALB rule priorities
The shared ALB seems to be consistently getting traffic in this order (by volume): Play, Exchange, CMS, ID, Castle, Remix

The current rule priority order is: Exchange, CMS, Play, ID, Publish, Castle

We could reduce the number of rules evaluations necessary for most requests by reordering the rules to match

farski:
Reorder the rules to optimize for current traffic patterns. Closes #359

It should avoid using any existing rule prefixes, even ones that are being unassigned.

### DIFF
--- a/stacks/root.yml
+++ b/stacks/root.yml
@@ -170,18 +170,20 @@ Mappings:
   # Higher-traffic apps should be given a higher priority (lower number) to
   # reduce the number of rules evaluations needed for more common requests
   ALBListenerRulePriorityMap:
+    Play:
+      prefix: 12
     Exchange:
       prefix: 13
     Cms:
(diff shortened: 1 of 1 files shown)


cavis:
Didn't "bad things" happen the last time we tried to reorder these via CFN?  Or maybe that had more to do with reusing priorities?

farski:
Hm, idk. What sort of bad things? Like it deployed and things broke?

cavis:
It either failed, or decided to destroy/recreate the load balancer.  I guess we can always try it in staging.

PR #369, opened 2019-11-24
Move Upload Lambda to Node 12

farski:


kookster:
I mean, the proof is in the pudding - guess we need to try this in staging, shouldn't have to worry about perf issues at least, given the infrequency with which this is called.

PR #371, opened 2019-12-06
Add email subscription to fatal SNS topic

### LINKED ISSUE #346: Fallback for CloudWatch Alarm notifications
Need some sort of fallback for situations where Lambda or Slack are not working, so alarms don't become invisible, especially for fatal alarms.

farski:
Closes #346

### DIFF
--- a/notifications/notifications.yml
+++ b/notifications/notifications.yml
@@ -91,6 +91,10 @@ Parameters:
     Type: String
     Description: >
       The name of the main CD pipeline in CodePipeline
+  FatalNotificationEmailAddress:
+    Type: String
+    Description: >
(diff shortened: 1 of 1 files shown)


notifications/notifications.yml
```diff
@@ -91,6 +91,10 @@ Parameters:
     Type: String
     Description: >
       The name of the main CD pipeline in CodePipeline
+  FatalNotificationEmailAddress:
```
cavis: So what will this actually be set to?

farski: It's deployed if you want to check. Not going to post the address because it's a new one and may as well stay private

PR #373, opened 2019-12-17
Updated postgres template

cavis:
Revamp the shared postgres template a bit (which ~i don't _think_ was~ was not in use for anything yet).

I'm using this to create new "shared" postgres databases in the data accounts, and port things over from legacy.


farski:
@cavis This has already been deployed to production?

cavis:
Yes.  So don't complain about it too much!

PR #376, opened 2020-01-24
Subscribe impressions to the ddb kinesis stream

svevang:
Sets up a subscription to the dovetail router log -- pushing those logs onto the bytes kinesis stream . Eventually, these dt-router impressions will be ingested by the bytes lambda as type 'ante-bytes' messages.

### DIFF
--- a/stacks/dovetail.prx.org.yml
+++ b/stacks/dovetail.prx.org.yml
@@ -51,9 +51,8 @@ Parameters:
     Type: String
   DovetailSecretsVersion:
     Type: String
-  DynamodbKinesisArn:
-    Type: String
-    default: Kinesis Stream Arn
+  DynamodbKinesisStream:
+    Type: AWS::SSM::Parameter::Value<String>
   # Dovetail Elixir ############################################################
   DovetailRouterEcrImageTag:
     Type: String
--- a/stacks/dovetail.prx.org.yml
+++ b/stacks/dovetail.prx.org.yml
@@ -52,7 +52,7 @@ Parameters:
   DovetailSecretsVersion:
     Type: String
   DynamodbKinesisArn:
-    type: String
+    Type: String
     default: Kinesis Stream Arn
   # Dovetail Elixir ############################################################
   DovetailRouterEcrImageTag:
--- a/stacks/dovetail.prx.org.yml
+++ b/stacks/dovetail.prx.org.yml
@@ -51,6 +51,9 @@ Parameters:
     Type: String
   DovetailSecretsVersion:
     Type: String
+  DynamodbKinesisArn:
+    type: String
+    default: Kinesis Stream Arn
   # Dovetail Elixir ############################################################
   DovetailRouterEcrImageTag:
     Type: String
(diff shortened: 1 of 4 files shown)


stacks/dovetail.prx.org.yml
```diff
@@ -51,6 +51,9 @@ Parameters:
     Type: String
   DovetailSecretsVersion:
     Type: String
+  DynamodbKinesisArn:
```
cavis: Okay, so here's what I'm thinking for getting this value into this stack (needs to resolve to `dovetail-ddb-staging` or `dovetail-ddb-production`)...

We already store the stream names in parameter-store.  You can actually see that happening with [serverless/analytics-ingest-lambdas.yml](https://github.com/PRX/Infrastructure/blob/master/stacks/serverless/analytics-ingest-lambdas.yml#L36) pulling the value, out of the key specified in the parent stack [serverless/root.yml](https://github.com/PRX/Infrastructure/blob/master/stacks/serverless/root.yml#L244).

So here, would need these param to have that AWS::SSM type, and the parent stack (root.yml) to interpolate and pass in the SSM param name.

stacks/dovetail.prx.org.yml
```diff
@@ -509,6 +512,46 @@ Resources:
           Value: !Ref AWS::StackName
         - Key: "prx:cloudformation:stack-id"
           Value: !Ref AWS::StackId
+  # connect dovetail-router impressions to ddb kinesis
+  DovetailRouterSubscriptionRole:
```
cavis: Can you add a `Condition: CreateStagingResources` to this for now?  Just in case!

stacks/dovetail.prx.org.yml
```diff
@@ -509,6 +512,46 @@ Resources:
...
+        - Key: "prx:cloudformation:stack-name"
+          Value: !Ref AWS::StackName
+        - Key: "prx:cloudformation:stack-id"
+          Value: !Ref AWS::StackId
+  DovetailRouterDynamodbKinesisSubscriptionFilter:
+    Type: "AWS::Logs::SubscriptionFilter"
+    Properties:
+      DestinationArn: !Ref DynamodbKinesisArn
+      FilterPattern: '{$.msg = impression}'
+      LogGroupName: !Ref DovetailRouterLogGroup.LogGroupName
```
cavis: Per the [CFN docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-loggroup.html) ... an `AWS::Logs::LogGroup` returns the "resource name" when you `!Ref DovetailRouterLogGroup` by itself.  Which is I think what we want here.  (And looking at other stacks in Infrastructure seems to confirm that).

stacks/dovetail.prx.org.yml
```diff
@@ -509,6 +512,46 @@ Resources:
...
+      RoleArn: !GetAtt DovetailRouterSubscriptionRole.arn
```
cavis: Should be capitalized `Arn`.  (Also, you can see these lint errors with [cfn-lint](https://github.com/aws-cloudformation/cfn-python-lint)... `cfn-lint stacks/dovetail.prx.org.yml --ignore-checks W`)

PR #380, opened 2020-03-25
CI stack SAM deploy

farski:
- Cleans up the CI template a bit (quote consistency, etc)
- Add the serverless transform
- Uses the serverless variant for Lambdas
- Switches the API to an HTTP (rather than REST) API Gateway

There shouldn't be any functional changes from this PR.


.vscode/tasks.json
```diff
@@ -0,0 +1,13 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=733558
+  // for the documentation about the tasks.json format
```
cavis, line 3: Comments in JSON?  That's blasphemy, vscode!

ci/ci.yml
```diff
@@ -150,25 +151,25 @@ Resources:
...
               - Effect: Allow
                 Action:
-                  - "ecr:*RepositoryPolicy"
-                  - "ecr:BatchCheckLayerAvailability"
-                  - "ecr:GetAuthorizationToken"
-                  - "ecr:InitiateLayerUpload"
-                  - "ecr:CompleteLayerUpload"
-                  - "ecr:UploadLayerPart"
-                  - "ecr:PutImage"
+                  - ecr:*RepositoryPolicy
```
cavis, line 159: Hmm... yeah, i guess unquoted `*` is fine as long as the string doesn't start with it.

ci/ci.yml
```diff
@@ -232,262 +233,71 @@ Resources:
...
-  CiStatusNotificationHandlerLambdaFunctionRetentionPolicy:
-    Type: "Custom::LogGroupRetentionPolicy"
-    Properties:
-      ServiceToken: !Ref LogGroupRetentionPolicyFunctionArn
-      LogGroupName: !Sub "/aws/lambda/${CiStatusNotificationHandlerLambdaFunction}"
-      RetentionInDays: 5
+        Project: Infrastructure
+        "prx:cloudformation:stack-name": !Ref AWS::StackName
+        "prx:cloudformation:stack-id": !Ref AWS::StackId
+  # HttpApiDomainNameRecordSetGroup:
```
cavis, line 259: Why commented out?  Just making sure the old recordset gets completely deleted first?

farski: Because of the way SAM transforms these SAM resources into actual Cfn resources behind the scenes, sometimes there are cases where you can't reference the same outputs you would if it were done in pure Cfn. It's kind of a limitation, and if this were a more important host name I would have just done it less magically, but for this I just switched the GitHub webhook to the domain provided by API Gateway.

farski:
Already deployed, so merging.

PR #382, opened 2020-03-25
DNS stack cert validation

farski:
The `prxu.org` DNS change has already been deployed.

The stack certificate change deploy should be watched somewhat closely. I have made sure that all the validation records it requires to mint the new certificate already exist, so it should deploy without intervention, but in case I missed something we'd want to create those records while it's deploying to prevent a rollback.

### DIFF
--- a/ci/ci.yml
+++ b/ci/ci.yml
@@ -235,9 +235,6 @@ Resources:
     Type: "AWS::CertificateManager::Certificate"
     Properties:
       DomainName: !Sub "*.${AWS::Region}.${AWS::StackName}.prx.tech"
-      DomainValidationOptions:
-        - ValidationDomain: prx.tech
-          DomainName: !Sub "*.${AWS::Region}.${AWS::StackName}.prx.tech"
       Tags:
         - Key: Project
           Value: Infrastructure
--- a/dns/prxu.org-hosted_zone.yml
+++ b/dns/prxu.org-hosted_zone.yml
@@ -218,6 +218,11 @@ Resources:
       HostedZoneId: !Ref HostedZone
       RecordSets:
         - ResourceRecords:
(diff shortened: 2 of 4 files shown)


ci/ci.yml
```diff
@@ -235,9 +235,6 @@ Resources:
     Type: "AWS::CertificateManager::Certificate"
     Properties:
       DomainName: !Sub "*.${AWS::Region}.${AWS::StackName}.prx.tech"
-      DomainValidationOptions:
-        - ValidationDomain: prx.tech
-          DomainName: !Sub "*.${AWS::Region}.${AWS::StackName}.prx.tech"
```
farski: This property was vestigial

etc/wildcard-certificate.yml
```diff
@@ -7,9 +7,6 @@ Resources:
     Type: AWS::CertificateManager::Certificate
     Properties:
       DomainName: "*.prx.tech"
-      DomainValidationOptions:
-        - ValidationDomain: prx.tech
-          DomainName: "*.prx.tech"
```
farski: This property was vestigial

dns/prxu.org-hosted_zone.yml
```diff
@@ -217,6 +217,11 @@ Resources:
       Comment: AWS ACM, etc
       HostedZoneId: !Ref HostedZone
       RecordSets:
+        - ResourceRecords:
+            - _f3ac86df105ba2b541203e3c55fc1a0b.acm-validations.aws.
+          TTL: "300"
+          Type: CNAME
+          Name: !Sub _4523c88a53cb676ba587436bd336533a.${Domain}
```
farski: Already deployed

PR #389, opened 2020-04-23
Dovetail Router basic alarms

cavis:
- [x] Add a target-group 5XX alarm for dovetail router specifically.  Dovetail classic already has some alarms, but they allow > 1 5XX.  For now, I want to see every 5XX dovetail-router sends.
- [x] Add an alarm on logged `{level: "error"}`s.
- [x] Alarm in both staging and prod, since we're testing out cloned traffic in staging right now.

### DIFF
--- a/stacks/dovetail.prx.org.yml
+++ b/stacks/dovetail.prx.org.yml
@@ -510,6 +510,53 @@ Resources:
           Value: !Ref AWS::StackName
         - Key: "prx:cloudformation:stack-id"
           Value: !Ref AWS::StackId
+  # Dovetail Elixir Alarms
+  DovetailRouterLogErrorsMetric:
+    Type: "AWS::Logs::MetricFilter"
+    Properties:
+      FilterPattern: '{ $.level = "error" }'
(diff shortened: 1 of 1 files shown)


stacks/dovetail.prx.org.yml
```diff
@@ -510,6 +510,53 @@ Resources:
...
+      LogGroupName: !Ref DovetailRouterLogGroup
+      MetricTransformations:
+        - MetricName: !Sub "dovetail_router_errors_${EnvironmentTypeAbbreviation}"
+          MetricNamespace: LogMetrics
+          MetricValue: "1"
+  DovetailRouterLogErrorsAlarm:
+    Type: "AWS::CloudWatch::Alarm"
+    Properties:
+      AlarmName: !Sub "[DovetailRouter][LogErrors] ${EnvironmentType}"
+      AlarmActions: [!Ref OpsErrorMessagesSnsTopicArn]
```
farski, line 527: We don't really do inline arrays in our templates. We could start I suppose… but it's non standard as things stand.

PR #390, opened 2020-04-24
Lambda@Edge support in single-origin CDN template

farski:
- Adds basic support for passing in the ARN for Lambdas to use a Edge handlers on the single origin template
- Adds a couple generic Lambdas to handle basic path rewriting, mainly for use with Feedburner feeds


farski:
This should be implemented differently if it's ever necessary

PR #391, opened 2020-04-26
Update CI Docker image version

farski:
V1 doesn't look to be supported any more, and is missing new runtimes, like node 12

### DIFF
--- a/ci/ci.yml
+++ b/ci/ci.yml
@@ -214,7 +214,7 @@ Resources:
             Value:
               Fn::ImportValue:
                 !Sub ${InfrastructureStorageStackName}-InfrastructureApplicationCodeBucket
-        Image: aws/codebuild/amazonlinux2-x86_64-standard:1.0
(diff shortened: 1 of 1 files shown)


ci/ci.yml
```diff
@@ -214,7 +214,7 @@ Resources:
             Value:
               Fn::ImportValue:
                 !Sub ${InfrastructureStorageStackName}-InfrastructureApplicationCodeBucket
-        Image: aws/codebuild/amazonlinux2-x86_64-standard:1.0
+        Image: aws/codebuild/amazonlinux2-x86_64-standard:3.0
```
cavis, line 217: But most of our repos run docker inside this codebuild container (docker in docker)... so this shouldn't really matter, right?  I can't think of any that run the tests directly in the parent container.

farski: Probably just my stuff, like Infrastructure and Porter

PR #394, opened 2020-05-06
Track how long ECS agents have been disconnected

cavis:
The ASG sizer lambda has been killing EC2 instances with disconnected ecs-agents a bit too frequently.  Turns out, sometimes the ECS `agentConnected` may briefly report `false` even though the instance is fine.

This encodes in a tag the epoch seconds when an EC2 instance is seen with an ECS agent disconnected.  Then the instance must remain disconnected for 5 minutes for to get manually terminated.

### DIFF
--- a/utility/lambdas/ecs-asg-sizer/lambda_function.py
+++ b/utility/lambdas/ecs-asg-sizer/lambda_function.py
@@ -16,7 +16,7 @@ ecs_client = session.client(service_name='ecs')
 sns_client = session.client(service_name='sns')
 
 EMPTY_SLOTS = 2
-ECS_AGENT_WAIT = 600
+ECS_AGENT_WAIT = 300
 SLACK_CHANNEL = '#ops-debug'
 SLACK_ICON = ':ops-autoscaling:'
 
@@ -24,7 +24,7 @@ SLACK_ICON = ':ops-autoscaling:'
 DISCONNECTS_TAG = 'ecs-asg-sizer:disconnects'
 DISCONNECTS_DELIMITER = ' '
 DISCONNECTS_SEPARATOR = ':'
-DISCONNECT_THRESHOLD_SECONDS = 600
+DISCONNECT_THRESHOLD_SECONDS = 300
 
 def lambda_handler(event, context):
     ASG_NAME = get_env('ASG_NAME')
--- a/utility/lambdas/ecs-asg-sizer/lambda_function.py
+++ b/utility/lambdas/ecs-asg-sizer/lambda_function.py
@@ -52,7 +52,7 @@ def lambda_handler(event, context):
             )
 
     # lookup cluster instances and remove disconnected agents
-    all_usages = get_usages(ECS_CLUSTER)
+    all_usages = get_usages(ECS_CLUSTER, DRY_RUN)
     usages = list(filter(lambda u: u['disconnectseconds'] < DISCONNECT_THRESHOLD_SECONDS, all_usages))
 
     # return early if already scaling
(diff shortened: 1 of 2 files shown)


utility/lambdas/ecs-asg-sizer/lambda_function.py
```diff
@@ -46,9 +51,9 @@ def slack(text, fields=None):
                 Message=json.dumps(msg),
             )
 
-    # lookup cluster instances, removing any disconnected agents
-    usages = get_usages(ECS_CLUSTER)
-    usages = list(filter(lambda u: u['connected'], usages))
+    # lookup cluster instances and remove disconnected agents
+    all_usages = get_usages(ECS_CLUSTER)
+    usages = list(filter(lambda u: u['disconnectseconds'] < DISCONNECT_THRESHOLD_SECONDS, all_usages))
```
svevang, line 56: Makes sense, filter those usages for a time greater than the threshold. Those are the gone-gone instances.

utility/lambdas/ecs-asg-sizer/lambda_function.py
```diff
@@ -210,22 +224,62 @@ def get_usages(cluster_name):
...
         cluster=cluster_name, containerInstances=instance_arns)
-    return list(map(get_usage, details['containerInstances']))
+    usages = list(map(get_usage, details['containerInstances']))
+
+    # mark elapsed time agents have been disconnected
+    cluster_arn = get_cluster_arn(cluster_name)
+    disconnects = get_disconnects(cluster_arn)
+    now = int(datetime.datetime.now().timestamp())
+    for usage in usages:
+        if usage['disconnected'] and disconnects.get(usage['id']):
```
svevang, line 234: Interesting, is there not an id in some cases?  Some properties of the cluster flickering?  I'm curious if this or other props indicate underlying state.  May be too hard to tell.

utility/lambdas/ecs-asg-sizer/lambda_function.py
```diff
@@ -210,22 +224,62 @@ def get_usages(cluster_name):
...
+                parts = value.split(DISCONNECTS_SEPARATOR)
+                if len(parts) == 2:
+                    disconnects[parts[0]] = int(parts[1])
+    return disconnects
+
+
+def set_disconnects(cluster_arn, previous_disconnects, usages, now):
+    disconnects = []
+    for usage in usages:
+        if usage['disconnected']:
```
svevang, line 275: Does this need the other side?  In the case where an instance comes back and then flickers off again -- having a stale `disconnectseconds` from the earlier flicker.  Maybe that never happens!

cavis: Yep - that should be covered.  The cases here are:

1. Instance disconnects for the 1st time - add to the `disconnects` hash.
2. Instance disconnects for a subsequent time - keep old timestamp in `disconnects` hash.
3. Instance _was_ disconnected, but this time around, it's connected.  Remove from the `disconnects` hash.  (Which happens implicitly because I'm rebuilding the list here).

I'm going to take another shot at organizing/commenting this method ... see if I can't make that clearer.

utility/lambdas/ecs-asg-sizer/lambda_function.py
cavis: I can do a better job commenting this as well.  The cases are:

1. This instance is connected.
2. This instance is disconnected for the first time ... so `disconnectseconds` is 0.
3. This instance is disconnected for the 2nd+ time ... need to calculate the elapsed `disconnectedseconds` using the current "now" epoch.

PR #400, opened 2020-05-14
Feat/ipv6 vpc v2

### LINKED ISSUE #88: Work on IPv6 support
https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-migrate-ipv6.html#vpc-migrate-assign-ipv6-address

- [ ] www.prx.org
- [ ] exchange.prx.org
- [x] play.prx.org
- [ ] publish.prx.org
- [ ] metrics.prx.org
- [ ] Dovetail
- [x] Dovetail CDN
- [ ] radiotopia.fm

farski:
Makes the following changes:
- [x] Remove the special case handling for creating subnets in `us-east-1`, which is no longer necessary
- [x] Determines a subnet's AZ using the native CloudFormation function `!GetAZs`, rather than being hardcoded
- [x] Determines the IPv4 and the new IPv6 CIDR blocks for a given subnet based on the CIDR block values returned by the VPC, rather than being hardcoded
-  [x] Adds an IPv6 CIDR block to the VPC
- [x] Adds a new route to the route table to handle IPv6 traffic
- [x] Duplicates all network ACL rules to handle IPv6 traffic
- [x] Update security groups to handle IPv6 traffic

Before this gets merged I think we need to come up with a reliable way of testing it.

Closes #88


stacks/vpc.yml
```diff
@@ -277,7 +304,9 @@ Resources:
           Value: !Ref AWS::StackName
         - Key: "prx:cloudformation:stack-id"
           Value: !Ref AWS::StackId
-  InboundHTTPSNetworkAclEntry:
+
+  # ACL inbound rules
```
farski, line 308: These rules are probably better to review outside the diff.

stacks/ecs.yml
```diff
@@ -366,6 +386,8 @@ Resources:
       SecurityGroupEgress:
         - CidrIp: 0.0.0.0/0
           IpProtocol: "-1"
+        - CidrIpv6: ::/0
+          IpProtocol: "-1"
```
cavis, line 390: Ah, didn't realize there was a wildcard for these.

stacks/vpc.yml
```diff
@@ -289,7 +326,20 @@ Resources:
       Protocol: 6 # TCP
       RuleAction : allow
       RuleNumber : 100
-  InboundHTTPNetworkAclEntry:
```
cavis, line 292: So you renamed all these ACL entries - what will that do, exactly?  Ideally the new ones will be created before the old ones drop, so we're not missing traffic.

farski: Yes, that's the CloudFormation way

farski:
@cavis Updated the rule numbers to avoid any collisions.

PR #416, opened 2020-06-24
CMS audio/image callback alarms

cavis:
Closes PRX/internal#543.  Deletes the cms audio/image lambdas and related resources, minus the SQS queues which are manually created.  Adds new alarms on SQS ... attempting to get a decent "you should wake up and fix this" fatal alarm.  Along with less noisy "it's just slow" alarms.

Eventually may want to convert this `cms-alarms.yml` stack into worker autoscaling rules.

Turns out, the old alarms are manually created, so after this merges will need to:

- [ ] Delete alarm `CMS audio-lambda SQS callbacks`
- [ ] Delete alarm `CMS image-lambda SQS callbacks`


stacks/container-apps/cms-alarms.yml
```diff
@@ -0,0 +1,115 @@
...
+      OKActions:
+        - !If [CreateProductionResources, !Ref OpsErrorMessagesSnsTopicArn, !Ref OpsWarnMessagesSnsTopicArn]
+      AlarmDescription: Audio callbacks are behind
+      ComparisonOperator: GreaterThanThreshold
+      EvaluationPeriods: 5
+      MetricName: ApproximateNumberOfMessagesVisible
+      Namespace: AWS/SQS
+      Period: 60
+      Statistic: Average
+      Threshold: 10
```
cavis, line 41: If we have > 10 callback messages waiting, every minute for 5 minutes ... that seems slow.  But not insane.

stacks/container-apps/cms-alarms.yml
```diff
@@ -0,0 +1,115 @@
...
+      OKActions:
+        - !Ref OpsFatalMessagesSnsTopicArn
+      AlarmDescription: Audio callbacks are maybe not processing
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      EvaluationPeriods: 15
+      MetricName: ApproximateNumberOfMessagesVisible
+      Namespace: AWS/SQS
+      Period: 60
+      Statistic: Average
+      Threshold: 1
```
cavis, line 88: Not sure this is the correct setting ... if CMS ever got busy enough, we could legitimately have 1 message waiting in the queue every minute for a long time.  But how to detect "there's only 1 job, but nothing is processing at all" then?

farski:
Oooh does this free up two root stack parameters? That's nice.

stacks/container-apps/cms-alarms.yml
```diff
@@ -0,0 +1,115 @@
...
+      OKActions:
+        - !If [CreateProductionResources, !Ref OpsErrorMessagesSnsTopicArn, !Ref OpsWarnMessagesSnsTopicArn]
+      AlarmDescription: Audio callbacks are behind
+      ComparisonOperator: GreaterThanThreshold
+      EvaluationPeriods: 3
+      MetricName: ApproximateAgeOfOldestMessage
+      Namespace: AWS/SQS
+      Period: 60
+      Statistic: Average
+      Threshold: 60
```
cavis, line 41: Okay, updated: so if there's any message > 1 minute old for 3 consecutive minutes ... error alarm.

stacks/container-apps/cms-alarms.yml
```diff
@@ -0,0 +1,115 @@
...
+      OKActions:
+        - !Ref OpsFatalMessagesSnsTopicArn
+      AlarmDescription: Audio callbacks are maybe not processing
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      EvaluationPeriods: 1
+      MetricName: ApproximateAgeOfOldestMessage
+      Namespace: AWS/SQS
+      Period: 60
+      Statistic: Average
+      Threshold: 300
```
cavis, line 88: And if _anything_ in the queue is older than 5 minutes ... fatal!

(Looking through the past month of data, we get a few minutes with datapoints slightly above 60, but nothing anywhere near 300).

PR #422, opened 2020-08-31
Set up a metric and alarm for augury

svevang:
Solves https://github.com/PRX/augury.prx.org/issues/156
Set up a metric and alarm for the augury workers to watch the invocations of workers pulling in actuals.

### DIFF
--- a/stacks/container-apps/augury-alarms.yml
+++ b/stacks/container-apps/augury-alarms.yml
@@ -38,7 +38,7 @@ Resources:
       AlarmDescription: Augury actuals are behind
       ComparisonOperator: LessThanThreshold
       EvaluationPeriods: 3
-      MetricName: AuguryActualsIngestJobInvocationMetric
+      MetricName: ActualsIngestJobInvocationMetric
       Namespace: AWS/ECS
       Period: 300
       Statistic: Sum
--- a/stacks/container-apps/augury-alarms.yml
+++ b/stacks/container-apps/augury-alarms.yml
@@ -21,7 +21,7 @@ Resources:
       FilterPattern: '[ImpressionsFetchJob] Success'
       LogGroupName: !Ref AuguryWorkerLogGroupName
       MetricTransformations:
-        - MetricName: !Sub "augury_actuals_rollups_ct_${EnvironmentTypeAbbreviation}"
+        - MetricName: !Sub "AuguryActualsRollupsCount${EnvironmentTypeAbbreviation}"
           MetricNamespace: PRX/Augury
           MetricValue: "1"
   AuguryActualsAreBehindAlarm:
--- a/stacks/container-apps/augury-alarms.yml
+++ b/stacks/container-apps/augury-alarms.yml
@@ -22,7 +22,7 @@ Resources:
       LogGroupName: !Ref AuguryWorkerLogGroupName
       MetricTransformations:
         - MetricName: !Sub "augury_actuals_rollups_ct_${EnvironmentTypeAbbreviation}"
-          MetricNamespace: LogMetrics
+          MetricNamespace: PRX/Augury
           MetricValue: "1"
   AuguryActualsAreBehindAlarm:
     Type: "AWS::CloudWatch::Alarm"
--- a/stacks/container-apps/augury-alarms.yml
+++ b/stacks/container-apps/augury-alarms.yml
@@ -15,13 +15,13 @@ Parameters:
   AuguryWorkerLogGroupName:
     Type: String
 Resources:
-  AuguryActualsIngestJobInvocationMetric:
+  ActualsIngestJobInvocationMetricFilter:
     Type: "AWS::Logs::MetricFilter"
     Properties:
       FilterPattern: '[ImpressionsFetchJob] Success'
       LogGroupName: !Ref AuguryWorkerLogGroupName
       MetricTransformations:
-        - MetricName: !Sub "AuguryActualsRollupsCount${EnvironmentTypeAbbreviation}"
+        - MetricName: !Sub "ActualsIngestJobInvocationCount${EnvironmentTypeAbbreviation}"
           MetricNamespace: PRX/Augury
           MetricValue: "1"
   AuguryActualsAreBehindAlarm:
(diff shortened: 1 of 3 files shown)


stacks/container-apps/augury-alarms.yml
```diff
@@ -0,0 +1,46 @@
...
+  OpsFatalMessagesSnsTopicArn:
+    Type: String
+  EnvironmentType:
+    Type: String
+  EnvironmentTypeAbbreviation:
+    Type: String
+  AuguryWorkerLogGroupName:
+    Type: String
+Resources:
+  AuguryActualsIngestJobInvocationMetric:
```
svevang, line 18: Couldn't find a good example of this in the webapps stack, Though the dovetail stack was somewhat helpful with the metric filters.  I ended up putting the metric right here next to the alarm. Is there a better spot?

stacks/container-apps/augury-alarms.yml
```diff
@@ -0,0 +1,46 @@
...
+      OKActions:
+        - !Ref OpsErrorMessagesSnsTopicArn
+      AlarmDescription: Augury actuals are behind
+      ComparisonOperator: LessThanThreshold
+      EvaluationPeriods: 3
+      MetricName: AuguryActualsIngestJobInvocationMetric
+      Namespace: AWS/ECS
+      Period: 300
+      Statistic: Sum
+      Threshold: 1
```
svevang, line 45: I'm reading this as: there has to be at least one invocation in 15 minutes or the alarm goes off.

stacks/container-apps/root.yml
```diff
@@ -364,6 +364,32 @@ Resources:
...
+            !Sub "${InfrastructureNotificationsStackName}-OpsWarnMessagesSnsTopicArn"
+        OpsErrorMessagesSnsTopicArn:
+          Fn::ImportValue:
+            !Sub "${InfrastructureNotificationsStackName}-OpsErrorMessagesSnsTopicArn"
+        OpsFatalMessagesSnsTopicArn:
+          Fn::ImportValue:
+            !Sub "${InfrastructureNotificationsStackName}-OpsFatalMessagesSnsTopicArn"
+        EnvironmentType: !Ref EnvironmentType
+        EnvironmentTypeAbbreviation: !Ref EnvironmentTypeAbbreviation
+        AuguryWorkerLogGroupName: !GetAtt AuguryStack.WorkerLogGroup
```
svevang, line 385: Got this to pass the linter.  I believe this resolves to the log group name, though may have to try and see.

Upon reflection, is this good way to pass this LogGroupName around?

stacks/container-apps/augury-alarms.yml
```diff
@@ -0,0 +1,46 @@
...
+    Type: String
+Resources:
+  AuguryActualsIngestJobInvocationMetric:
+    Type: "AWS::Logs::MetricFilter"
+    Properties:
+      FilterPattern: '[ImpressionsFetchJob] Success'
+      LogGroupName: !Ref AuguryWorkerLogGroupName
+      MetricTransformations:
+        - MetricName: !Sub "augury_actuals_rollups_ct_${EnvironmentTypeAbbreviation}"
+          MetricNamespace: LogMetrics
```
farski, line 25: I think it'd be better to put these metrics into their own namespace. `PRX/Augury` would be my vote.

stacks/container-apps/augury-alarms.yml
```diff
@@ -0,0 +1,46 @@
...
+  AuguryWorkerLogGroupName:
+    Type: String
+Resources:
+  AuguryActualsIngestJobInvocationMetric:
+    Type: "AWS::Logs::MetricFilter"
+    Properties:
+      FilterPattern: '[ImpressionsFetchJob] Success'
+      LogGroupName: !Ref AuguryWorkerLogGroupName
+      MetricTransformations:
+        - MetricName: !Sub "augury_actuals_rollups_ct_${EnvironmentTypeAbbreviation}"
```
farski, line 24: If these metrics get moved to their own namespace, maybe the metric name can be tweaked a bit. Also, by convention, CloudWatch metrics are named using PascalCase; the Dovetail ones are a little unusual in that regard.

stacks/container-apps/augury-alarms.yml
```diff
@@ -0,0 +1,46 @@
...
+    Type: String
+  EnvironmentTypeAbbreviation:
+    Type: String
+  AuguryWorkerLogGroupName:
+    Type: String
+Resources:
+  AuguryActualsIngestJobInvocationMetric:
+    Type: "AWS::Logs::MetricFilter"
+    Properties:
+      FilterPattern: '[ImpressionsFetchJob] Success'
```
farski, line 21: Are there any other patterns to track? Failures, maybe? Or any specific values coming out of each job?

stacks/container-apps/augury-alarms.yml
svevang: Ahh, right.  So I think 5 minutes is fine here, just was incorporating the `EvaluationPeriods` in that overall view.

Rephrasing I imagine this to work: if in 3 consecutive periods of 5 minutes each, if there are no matching log lines (indicating a successful actuals-ingest job completion) then the alarm goes off.

stacks/container-apps/augury-alarms.yml
svevang: Any particular tweaking to the name? Maybe add the namespace as a prefix?

farski (changes requested):

stacks/container-apps/augury-alarms.yml
```diff
@@ -0,0 +1,46 @@
...
+      AlarmActions:
+        - !Ref OpsErrorMessagesSnsTopicArn
+      InsufficientDataActions:
+        - !Ref OpsErrorMessagesSnsTopicArn
+      OKActions:
+        - !Ref OpsErrorMessagesSnsTopicArn
+      AlarmDescription: Augury actuals are behind
+      ComparisonOperator: LessThanThreshold
+      EvaluationPeriods: 3
+      MetricName: ActualsIngestJobInvocationMetric
```
farski, line 41: This no longer matches the metric name on L24

farski (changes requested):

cavis (commented):
+1 to normalizing how we namespace metric filters going forward.  Sounds like we'll deprecate the LogMetrics namespace, instead using an app-specific namespace.  But continue to put the Environment name in the MetricName.

Yeah, what farski said about metric names not matching in here.  But otherwise looks good to me!

farski:
Yeah, until they support dimensioning filter metrics, we'll have to just put the environment in the name

farski (approved):

PR #423, opened 2020-09-01
Remove legacy WordPress DNS records

farski:
Continues the work on https://github.com/PRX/internal/issues/520


cavis (approved):
👍  looks good to me

PR #424, opened 2020-09-10
Add m5 instance sizes, lint Cfn template

kookster:
Added in newer m5 instance sizes.


db/mysql-instance.yml
```diff
@@ -69,6 +69,12 @@ Parameters:
       - db.m4.2xlarge
       - db.m4.4xlarge
       - db.m4.10xlarge
+      - db.m5.large
```
kookster, line 72: added instances from here: https://aws.amazon.com/rds/mysql/instance-types/

farski (approved):

PR #426, opened 2020-09-11
Upgrade all HTTP requests on platform ALB

farski:
Should we decide we want to enforce HTTPS on all our apps (handled by this ALB), this rule should do that.

Once we're good with the change in production, it would be better to move this to the default HTTP listener rule


farski:
@cavis Thoughts on this?

cavis:
Which apps are impacted?  We've got ID and all our SPAs already upgrading themselves ... so that leaves Exchange, CMS, Augury.  Others?

Might be worth auditing some of our inter-app communications to make sure they're all using HTTPS.  I'm pretty sure there are some cases where Dovetail makes an HTTP call out to Feeder (though that is a different ALB).

farski:
- Play
- Exchange
- CMS
- ID
- Castle
- Remix
- Publish
- Metrics 
- Grove
- Crier
- Stripe Proxy
- Courier
- Augury
- Styleguide
- Networks
- Iframely

cavis (approved):
👍 i'm fine with trying this in staging.

PR #429, opened 2020-09-29
Remove Stripe Proxy

farski:
This also removes the LegacySupport stack since Stripe Proxy was the only thing in there, and templates without resources can't be deployed.

Closes https://github.com/PRX/internal/issues/588


stacks/container-apps/sphinx-nlb.yml
```diff
@@ -474,7 +474,7 @@ Resources:
           TargetGroupArn: !Ref NLBSphinxTargetGroup
       # Role: !Ref ECSServiceIAMRole
       NetworkConfiguration:
-        AwsvpcConfiguration:
+        AwsVpcConfiguration:
```
farski, line 477: This either changed, or cfn-lint was allowing both and now isn't.

cavis (approved):
👍

stacks/container-apps/sphinx-nlb.yml
kookster: That seems to be right now though
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-service-networkconfiguration.html

kookster (approved):
lgtm

PR #431, opened 2020-10-05
Modernize CI Lambdas

farski:
In preparation for [the upcoming changes](https://github.com/github/renaming/) to default branch naming, this will allow the CI system to build properly for any default branch name, not just `master`.

This PR also modernizes the Lambda code for several steps in the CI process. Key changes are: replacing some callbacks and `then` calls with asycn/await; adding type hinting; replacing several uses of a deprecated url method.


farski:
I'd like to test this a bit before it gets merged

sandikbarr (approved):
👍  makes good sense, lgtm

PR #432, opened 2020-10-06
Replace deprecated cache properties in single-origin template

farski:
Replaces some deprecated properties (DefaultTTL, MaxTTL, MinTTL, ForwardedValued) with new CachePolicy and OriginRequestPolicy resources.

This allows for more granular control over differences between caching behavior and origin request behavior. It also allows support for newer features, like Brotli compression.

Builds will fail until cfn-lint catches up on Brotli support.

### DIFF
--- a/cdn/single-origin.yml
+++ b/cdn/single-origin.yml
@@ -190,35 +190,39 @@ Parameters:
     Type: String
     Description: >-
       (optional) The default amount of time, in seconds, that you want objects
-      to stay in CloudFront caches before CloudFront forwards another request
-      to your origin to determine whether the object has been updated. The
-      value that you specify applies only when your origin does not add HTTP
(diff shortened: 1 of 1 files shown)


cavis (approved):
👍 looks right to me.

Is the plan to update existing distributions then?  Looks like you can switch between them in the CloudFront web UI ... so you'd think it would just work, since you kept the input params the same.

farski:
I'm not going to go out of my way to redeploy every CDN we have. There may be a few that could particularly benefit from the Brotli compression, but as things eventually get redeployed my expectation is that this version of the template is entirely backwards compatible with the old one. I wouldn't think this would be the type of thing that CloudFormation can do in a single deploy; that wouldn't be great.

cavis:
Makes sense.  Good sense!

farski:
Going to try a deploy before merging

PR #435, opened 2020-10-12
Enable prod networks and sphinx

kookster:
Time to try deploying networks to prod!
This gets rid of the IsStaging conditionals, I think that's all that needs to be done?
(secrets are already setup and deployed)


farski (approved):
👍 🎉

PR #444, opened 2020-10-26
Select Mysql EngineVersion from most recent supported versions

kookster:
added from here:
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_MySQL.html#MySQL.Concepts.VersionMgmt


cavis (approved):
👍

PR #446, opened 2020-10-27
ECS task IAM role testing

farski:
Attempting to add a IAM role that only applies to the Courier web worker. I'm not intending this to every be done this way for a real app in production, just need to have something deployed to start checking behavior.


stacks/container-apps/web-application.yml
```diff
@@ -230,6 +235,76 @@ Resources:
     Type: "AWS::Logs::LogGroup"
     Properties:
       RetentionInDays: 14
+  CourierWebTaskIamRole:
```
cavis, line 238: Oof, parameterizing roles seems like a pain.  Maybe we'll just store these in another stack and import them?  (Just explicitly define them for every app).

cavis (approved):
I'm curious if you'll need AWS-secrets get/decode permissions.  But +1 on this experiment!

PR #448, opened 2020-10-27
Make the Mysql RDS template more similar to the Postgresql template

kookster:
Still some more work to make them equivalent, but this is closer

### DIFF
--- a/db/mysql-instance.yml
+++ b/db/mysql-instance.yml
@@ -1,7 +1,7 @@
 # db/mysql-instance.yml
 AWSTemplateFormatVersion: "2010-09-09"
 Description: >-
-  AWS CloudFormation RDS Template: Template to create RDS MySQL v5.5 instance.
+  AWS CloudFormation RDS Template: Template to create RDS MySQL instance.
 Parameters:
   DBInstance:
     Default: "name-me"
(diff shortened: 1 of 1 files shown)


db/mysql-instance.yml
```diff
@@ -162,35 +173,41 @@ Resources:
...
+      DeleteAutomatedBackups: true
+      DeletionProtection: true
+      EnablePerformanceInsights: !If [IsProduction, true, false]
       Engine: MySQL
       EngineVersion: !Ref EngineVersion
       MasterUsername: !If [UseSnapshot, !Ref "AWS::NoValue", !Ref DBUser]
       MasterUserPassword:
         !If [UseSnapshot, !Ref "AWS::NoValue", !Ref DBPassword]
-      MultiAZ: false
+      MultiAZ: !If [IsProduction, true, false]
```
farski, line 196: This seems like a significant change for existing DBs being updated from the previous version of the template for production environments

db/mysql-instance.yml
```diff
@@ -162,35 +173,41 @@ Resources:
...
+      PerformanceInsightsRetentionPeriod: !If [IsProduction, 7, !Ref "AWS::NoValue"]
+      PreferredBackupWindow: "08:00-08:30"
+      PreferredMaintenanceWindow: "tue:06:15-tue:06:45"
+      PubliclyAccessible: true
```
farski, line 200: This also presents a potentially significant change for updating resources from the previous version of the template

kookster:
superseded by #453

PR #449, opened 2020-10-28
Modernize storage template

farski:
- Adds UpdateReplacePolicy for all bucket resources
- Replaces a `join` to build the bucket URL with a more direct return value, which didn't exist when the template was first made. (This does change the URL from a path-style URL to virtual host. I don't think this value is actually being used anywhere, but it shouldn't matter even if it is)
- Small syntax changes


cavis (approved):
👍 looks good to me!

PR #451, opened 2020-11-09
Pull dockerhub login from parameterstore

cavis:
Note: this would save a couple lines, but each individual repo's buildspec would still need to run the [docker login](https://github.com/PRX/exchange.prx.org/commit/c016a6f500644b44b492d83371e818354f659c0f#diff-69387ac97f1b775f19989fc28150f89e785406cbff04643a969f5e396573dd5cR14) command itself.


farski:
I think we've handled this in the app-specific repos, closing

PR #452, opened 2020-11-11
Update ECS AMI and change instance type in staging

farski:


cavis (approved):
Looks good to me.  Double-checked the [ecs-agent CHANGELOG](https://github.com/aws/amazon-ecs-agent/blob/master/CHANGELOG.md) for 1.39.0 to 1.47.0.  All these changes look good.

I'm good with trying out the `t3a` instances, but should maybe double check our CPU usages over the past couple months.  Just to make sure we don't spike to 100% for long periods of time ... which sounds like the main downside to the AMD instances.

PR #453, opened 2020-11-11
Copy/update postgres template for mysql

cavis:
Copy the postgres template over to mysql, and edit from there (for specific mysql stuff).

Used to launch the prod shared mysql database.


farski (approved):
The proof is in the pudding, I guess

PR #454, opened 2020-11-12
Add HTTP API and SAM function for Pixel tracker

farski:
Adds an unused HTTP API version of the Pixel API alongside the REST API, to do some initial testing.

This is to start the process of moving to faster/cheaper HTTP APIs in cases where we don't rely on any REST-only functionality.


cavis (approved):
So just spinning up a parallel function/API?  +1 seems fine to me.

PR #457, opened 2020-11-17
Update some node Lambdas

farski:
Updates the Lambda Node version used for the proxy app and pixel tracker


kookster (approved):
short and sweet - lgtm

PR #458, opened 2020-11-18
Modernize the play.prx.org CloudFront resources

farski:
- Enable IPv6
- Switch to cache/request policies
- Enable brotli compression
- Remove unnecessary comments

### DIFF
--- a/stacks/play.prx.org.yml
+++ b/stacks/play.prx.org.yml
@@ -212,6 +212,40 @@ Resources:
             DNSName: !GetAtt PlayCloudFrontDistribution.DomainName
             HostedZoneId: Z2FDTNDATAQYW2
   # CloudFront
+  PlayCloudFrontCachePolicy:
+    Type: AWS::CloudFront::CachePolicy
+    Properties:
+      CachePolicyConfig:
+        Comment: !Sub Policy for ${AWS::StackName}
+        DefaultTTL: 86400
+        MaxTTL: 31536000
+        MinTTL: 0
+        Name: !Sub ${AWS::StackName}-CachePolicy
+        ParametersInCacheKeyAndForwardedToOrigin:
+          CookiesConfig:
+            CookieBehavior: none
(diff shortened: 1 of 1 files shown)


stacks/play.prx.org.yml
```diff
@@ -212,6 +212,40 @@ Resources:
...
+        Comment: !Sub Policy for ${AWS::StackName}
+        CookiesConfig:
+          CookieBehavior: none
+        HeadersConfig:
+          HeaderBehavior: none
+        Name: !Sub ${AWS::StackName}-OriginRequestPolicy
+        QueryStringsConfig:
+          QueryStringBehavior: whitelist
+          QueryStrings:
+            - url
```
cavis, line 248: Just thinking this through (even though it's the same setting we had before) ...

This will cache any RSS proxied by the express server.  But other pages with query params (the embed player angular app itself) ... it will just cache 1 copy of that HTML.  But that's fine, since all the HTML is the same, and js/angular is parsing the query params itself.

stacks/play.prx.org.yml
```diff
@@ -212,6 +212,40 @@ Resources:
             DNSName: !GetAtt PlayCloudFrontDistribution.DomainName
             HostedZoneId: Z2FDTNDATAQYW2
   # CloudFront
+  PlayCloudFrontCachePolicy:
+    Type: AWS::CloudFront::CachePolicy
+    Properties:
+      CachePolicyConfig:
+        Comment: !Sub Policy for ${AWS::StackName}
+        DefaultTTL: 86400
```
cavis, line 220: These are the defaults, right?  So not actually changing anything.  And I think all the play.prx.org endpoints set their own TTLs anyways.

cavis (approved):
👍  looks good to me

stacks/play.prx.org.yml
farski: They are the defaults but for some reason those properties are required in the new format

PR #463, opened 2020-12-04
Add support for a failover origin to the single origin template

farski:
This is a backwards-compatible update to the template that allows for an optional failover domain to be passed in as a parameter. When that parameter is present, an additional origin is created, both origins are placed into an _origin group_, and the origin group is set as the target for the default cache behavior. When the parameter is omitted, the template behaves as it did prior to this change.

The failover configuration supports any of the following status codes acting as failover triggers: 403, 404, 500, 502, 503, 504. Currently the template offers three sets of values. Any other combination of codes can be added should the need arise.

### DIFF
--- a/cdn/single-origin.yml
+++ b/cdn/single-origin.yml
@@ -25,7 +25,9 @@
 # }
 AWSTemplateFormatVersion: "2010-09-09"
 Description: >-
-  Creates a CloudFront distribution with a single origin (either web or S3)
+  Creates a CloudFront distribution with a single origin (either web or S3) and
+  an optional failover origin. The primary and failover origin share an
+  identical configuration, aside from the domain.
 Conditions:
   HasCloudFrontDefaultRootObject: !Not [!Equals [!Ref CloudFrontDefaultRootObject, ""]]
   HasCloudFrontCompress: !Equals [!Ref CloudFrontCompress, "True"]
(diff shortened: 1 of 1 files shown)


cdn/single-origin.yml
```diff
@@ -153,11 +173,34 @@ Parameters:
...
+    Type: String
+    Description: >-
+      The status codes that, when returned by the primary origin, trigger
+      CloudFront to attempt to use the failover origin.
+    Default: "403-404-500-502-503-504"
+    AllowedValues:
+      # When adding an additional value, a matching entry must be added to the
+      # FailoverCriteriaStatusCodesQuantity mapping which defines the number of
+      # codes present in the value.
+      - "403-404"
```
cavis, line 201: Does this happen before or after status codes are translated by the CDN?  (We usually change 403s to 404s in our distributions... since 403 is what S3 actually returns when an object doesn't exist).

I suppose if you've always got 403/404 grouped together here, it doesn't actually matter.

cdn/single-origin.yml
```diff
@@ -436,7 +499,7 @@ Resources:
                     - TLSv1.1
                     - TLSv1
             DomainName: !Ref OriginDomain
-            Id: web-origin
+            Id: primary-origin
```
cavis, line 502: Will this tear down the origin for existing stacks?  What impact does that have on cached objects?  Hopefully they wouldn't get invalidated because they came from a gone origin.

cavis (approved):
👍 Couple questions, but i think this looks good.  And seems like you've already tested it too.

cdn/single-origin.yml
farski: > The primary origin returns an HTTP status code that you’ve configured for failover

I believe failover is based on the actual response code coming back from the origins

farski: Also worth noting:

>CloudFront routes all incoming requests to the primary origin, even when a previous request failed over to the secondary origin. CloudFront only sends requests to the secondary origin after a request to the primary origin fails.

cdn/single-origin.yml
farski: I can't say for sure. I believe this will create a new origin record under the distribution, because I don't see anyway to set an origin ID other than when creating it (i.e., they're immutable). I don't think that would bust any caches, though. My understanding is that object key evaluation happens first within the edge, and if there's a miss only then does CloudFront attempt to get the resource from an origin.

farski: I also don't know that we have any distributions based on this template that would be particularly hard-hit even if the global cache got busted. We can be safer with something like that DT bucket if/when we implement the origin groups.

PR #468, opened 2021-01-26
Classy CNAME for support.prx.org

klbaradan:
CNAME for support.prx.org to vanity.classy.org

### DIFF
--- a/dns/prx.org-hosted_zone.yml
+++ b/dns/prx.org-hosted_zone.yml
@@ -1571,4 +1571,4 @@ Resources:
             - vanity.classy.org.
           TTL: "300"
           Type: CNAME
-          Name: !Sub support.${Domain}
+          Name: !Sub give.${Domain}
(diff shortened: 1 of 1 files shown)


dns/prx.org-hosted_zone.yml
```diff
@@ -1561,3 +1561,14 @@ Resources:
...
+    Type: "AWS::Route53::RecordSetGroup"
+    Properties:
+      Comment: Classy
+      HostedZoneId: !Ref HostedZone
+      RecordSets:
+        - ResourceRecords:
+            - vanity.classy.org.
+          TTL: "300"
+          Type: CNAME
+          Name: !Sub support.${Domain}
```
farski, line 1574: Just to confirm, we're assigning all of `support.prx.org` to Classy?

klbaradan: That was the intent, but ... now that you ask, it's at least worth considering that support can have many meanings, doubly so for us... Will run it by the (Resource) Development team!

klbaradan: OK we're going to go with give instead, thanks for catching that!!

farski (approved):

PR #469, opened 2021-01-26
Publish S3 bucket refactor

farski:
Creates a new bucket to be used (in the future) for Publish uploads temp storage. For now it just passes the environment variables in in a way that matches how they will work with the new bucket.


cavis (approved):
👍

PR #472, opened 2021-02-09
Use inline Publish bucket and CDN in production

### LINKED ISSUE #471: Configure Publish to access uploads via CloudFront
- [ ] Add a CloudFront distribution to the Publish stack with the upload bucket as the origin
- [ ] Pass the CDN domain to Publish as the `publicAccessHost`
- [ ] Confirm that uploads are being requested via the CDN after they've been uploaded but haven't been saved+processed yet

farski:
Closes #471

Flips the switch on using the new template-provided S3 bucket for Uploads, as well as a CloudFront distribution in front of the bucket to provide the necessary public access to the uploads


cavis (approved):
👍  all seems to be working in staging.  Onwards!

PR #476, opened 2021-02-17
Dovetail cdn arranger lambda

cavis:
Init the dovetail-cdn-arranger deployment.

### DIFF
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -13,3 +13,6 @@ overrides:
     options:
       tabWidth: 4
       useTabs: true
+  - files: '*.yml'
+    options:
+      singleQuote: false
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -1,18 +1,18 @@
 semi: true
-trailingComma: 'all'
+trailingComma: "all"
 singleQuote: true
 printWidth: 80
 useTabs: false
 tabWidth: 2
 endOfLine: lf
 overrides:
-  - files: '*.md'
+  - files: "*.md"
     options:
       tabWidth: 4
-  - files: 'Makefile'
+  - files: "Makefile"
     options:
       tabWidth: 4
       useTabs: true
-  - files: '*.yml'
+  - files: "*.yml"
     options:
       singleQuote: false
--- a/stacks/serverless/dovetail-cdn-arranger.yml
+++ b/stacks/serverless/dovetail-cdn-arranger.yml
@@ -1,5 +1,5 @@
 # stacks/serverless/dovetail-cdn-arranger.yml
-AWSTemplateFormatVersion: 2010-09-09
+AWSTemplateFormatVersion: '2010-09-09'
 Description: Dovetail CDN Arranger lambda function
 Conditions:
   IsProduction: !Equals [!Ref EnvironmentType, Production]
@@ -48,7 +48,7 @@ Resources:
     Type: AWS::IAM::Role
     Properties:
       AssumeRolePolicyDocument:
-        Version: 2012-10-17
+        Version: '2012-10-17'
         Statement:
           - Effect: Allow
             Principal:
@@ -60,7 +60,7 @@ Resources:
       Policies:
         - PolicyName: DynamoDBWritePolicy
           PolicyDocument:
-            Version: 2012-10-17
+            Version: '2012-10-17'
             Statement:
               - Effect: Allow
                 Action:
@@ -71,7 +71,7 @@ Resources:
                   - !Ref DynamodbTableArn
         - PolicyName: S3WritePolicy
           PolicyDocument:
-            Version: 2012-10-17
+            Version: '2012-10-17'
             Statement:
               - Effect: Allow
                 Action:
--- a/stacks/serverless/dovetail-cdn-arranger.yml
+++ b/stacks/serverless/dovetail-cdn-arranger.yml
@@ -19,6 +19,8 @@ Parameters:
 Resources:
   DovetailCdnArrangerWorkspaceBucket:
     Type: AWS::S3::Bucket
+    DeletionPolicy: Retain
+    UpdateReplacePolicy: Retain
     Properties:
       LifecycleConfiguration:
         Rules:
--- a/stacks/serverless/dovetail-cdn-arranger.yml
+++ b/stacks/serverless/dovetail-cdn-arranger.yml
@@ -1,5 +1,5 @@
 # stacks/serverless/dovetail-cdn-arranger.yml
-AWSTemplateFormatVersion: '2010-09-09'
+AWSTemplateFormatVersion: "2010-09-09"
 Description: Dovetail CDN Arranger lambda function
 Conditions:
   IsProduction: !Equals [!Ref EnvironmentType, Production]
@@ -49,7 +49,7 @@ Resources:
     Type: AWS::IAM::Role
     Properties:
       AssumeRolePolicyDocument:
-        Version: '2012-10-17'
+        Version: "2012-10-17"
         Statement:
           - Effect: Allow
             Principal:
@@ -61,7 +61,7 @@ Resources:
       Policies:
         - PolicyName: DynamoDBWritePolicy
           PolicyDocument:
-            Version: '2012-10-17'
+            Version: "2012-10-17"
             Statement:
               - Effect: Allow
                 Action:
@@ -72,7 +72,7 @@ Resources:
                   - !Sub arn:aws:dynamodb:*:*:table/${ArrangementsDynamodbTable}
         - PolicyName: S3WritePolicy
           PolicyDocument:
-            Version: '2012-10-17'
+            Version: "2012-10-17"
             Statement:
               - Effect: Allow
                 Action:
--- a/stacks/serverless/dovetail-cdn-arranger.yml
+++ b/stacks/serverless/dovetail-cdn-arranger.yml
@@ -1,5 +1,5 @@
 # stacks/serverless/dovetail-cdn-arranger.yml
-AWSTemplateFormatVersion: '2010-09-09'
+AWSTemplateFormatVersion: 2010-09-09
 Description: Dovetail CDN Arranger lambda function
 Conditions:
   IsProduction: !Equals [!Ref EnvironmentType, Production]
@@ -20,7 +20,6 @@ Resources:
   DovetailCdnArrangerWorkspaceBucket:
     Type: AWS::S3::Bucket
     Properties:
-      # TODO: tag-specific expirations
       LifecycleConfiguration:
         Rules:
           - ExpirationInDays: 30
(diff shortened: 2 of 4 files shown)


stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+    Type: String
+  CodeS3ObjectVersion:
+    Type: String
+  EnvironmentType:
+    Type: String
+  DovetailHost:
+    Type: AWS::SSM::Parameter::Value<String>
+  DovetailToken:
+    Type: AWS::SSM::Parameter::Value<String>
+  DynamodbTableArn:
```
cavis, line 17: I still have some questions about how the nodejs code gets access to this.

- Should this be a table _name_ or _arn_?
- Is this table even in this account/region?
- Can we create a write-permission on this if it's not part of this account/region?

I seem to recall jumping through a bunch of hoops in the analytics lambda to get access to a data-production table.  Assuming a role linked to the other account and whatnot.  Wondering if there's an easier way here...

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+    Type: AWS::SSM::Parameter::Value<String>
+Resources:
+  DovetailCdnArrangerWorkspaceBucket:
+    Type: AWS::S3::Bucket
```
farski, line 21: This should have a `DeletionPolicy` and `UpdateReplacePolicy` set to `Retain` even though we don't actually want to retain the buckets or data. Because it's unlikely this bucket would exist and be empty, not having the policies would mess with tearing down a stack.

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+    Properties:
+      # TODO: tag-specific expirations
```
farski, line 23: I feel like tag-based lifecycles it part of what got us into trouble with the `prx-dovetail` bucket. I would prefer using multiple buckets if we find that this app is managing files with very distinct lifecycle requirements. It also really simplifies cost tracking to have things in different buckets.

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+        - Key: prx:cloudformation:stack-name
+          Value: !Ref AWS::StackName
+        - Key: prx:cloudformation:stack-id
+          Value: !Ref AWS::StackId
+  DovetailCdnArrangerBinaryLayer:
+    Type: AWS::Lambda::LayerVersion
+    Properties:
+      CompatibleRuntimes:
+        - nodejs14.x
+        - nodejs12.x
```
farski, line 42: Do we need 12.x compatibility?

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+      Content:
+        S3Bucket: prx-tech
+        S3Key: archives/ffmpeg-4.3.1-lambda-layer.zip
+      Description: ffmpeg binaries for dovetail-cdn-arranger
+      LicenseInfo: MIT
+  DovetailCdnArrangerExecutionIAMRole:
+    Type: 'AWS::IAM::Role'
```
farski, line 49: Prefer no quotes here

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+          - Effect: 'Allow'
```
farski, line 54: No quotes

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+            Principal:
+              Service:
+                - 'lambda.amazonaws.com'
```
farski, line 57: No quotes

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+            Action:
+              - 'sts:AssumeRole'
```
farski, line 59: No quotes

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+      Path: '/'
+      Policies:
+        - PolicyName: DynamoDBWritePolicy
+          PolicyDocument:
+            Version: '2012-10-17'
```
farski, line 64: Prefer double quotes

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+              - Effect: Allow
+                Action:
+                  - 'dynamodb:DescribeTable'
+                  - 'dynamodb:GetItem'
+                  - 'dynamodb:PutItem'
+                Resource:
+                  - !Ref DynamodbTableArn
+        - PolicyName: S3WritePolicy
+          PolicyDocument:
+            Version: '2012-10-17'
```
farski, line 75: Prefer double quotes

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
+# stacks/serverless/dovetail-cdn-arranger.yml
+AWSTemplateFormatVersion: '2010-09-09'
```
farski, line 2: Prefer double quotes

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+      Content:
+        S3Bucket: prx-tech
+        S3Key: archives/ffmpeg-4.3.1-lambda-layer.zip
+      Description: ffmpeg binaries for dovetail-cdn-arranger
+      LicenseInfo: MIT
+  DovetailCdnArrangerExecutionIAMRole:
+    Type: 'AWS::IAM::Role'
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: '2012-10-17'
```
farski, line 52: Prefer double quotes

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+          S3_BUCKET: !Ref DovetailCdnArrangerWorkspaceBucket
+      Handler: index.handler
+      Layers: !If
+        - IsProduction
+        - - !Sub arn:aws:lambda:${AWS::Region}:580247275435:layer:LambdaInsightsExtension:12
+        - !Ref AWS::NoValue
+      MemorySize: 8000
+      Role: !GetAtt DovetailCdnArrangerExecutionIAMRole.Arn
+      Runtime: nodejs14.x
+      Timeout: 300
```
farski, line 122: Any reason not to go with 15 minutes?

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+      Tags:
+        - Key: Project
+          Value: Dovetail
```
farski, line 125: This tag is going to get less and less useful over time as we consider more and more things Dovetail. I think it may be time to add an additional, more specific cost allocation tag. Something like `Application=dovetail-cdn-arranger`.

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+        - Key: Environment
+          Value: !Ref EnvironmentType
+        - Key: Name
+          Value: !Sub Dovetail-Counts-${EnvironmentType}
```
farski, line 129: Not Counts

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+      Code:
+        S3Bucket: !Ref CodeS3Bucket
+        S3Key: lambda/PRX-dovetail-cdn-arranger.zip
+        S3ObjectVersion: !Ref CodeS3ObjectVersion
+      Description: Dovetail CDN Arranger
+      Environment:
+        Variables:
+          DDB_TABLE: !Ref DynamodbTableArn
+          DOVETAIL_HOST: !Ref DovetailHost
+          DOVETAIL_TOKEN: !Ref DovetailToken
```
farski, line 112: Without knowing specifically what `host` and `token` are here, I feel like they may be the type of values that would be good to have version history on, which we do kind of get with Parameter Store, but not in a way that is compatible with our rollbacks. In the absence of something more similar to `aws-secrets` for Lambdas, I'm wondering if we move these to the JSON file.

stacks/serverless/dovetail-cdn-arranger.yml
farski, line 17: Could this parameter name be more descriptive of what the table contains?

farski: I've only ever seen the JavaScript SDK expecting a DDB table **name**, so that is what I generally pass to Lambda functions in the `Environment`.

Where the table is shouldn't matter too much. IAM never cares about regions for access. For access across accounts, assuming a role with a trust policy is the correct way of doing it.

As for where the table lives, based on what I remember this table holds data that we would consider important and not want to lose, and probably even be a global table. That to me feels like it would live in the data-prod account

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+    Type: 'AWS::Lambda::Function'
+    Properties:
+      Code:
+        S3Bucket: !Ref CodeS3Bucket
+        S3Key: lambda/PRX-dovetail-cdn-arranger.zip
+        S3ObjectVersion: !Ref CodeS3ObjectVersion
+      Description: Dovetail CDN Arranger
+      Environment:
+        Variables:
+          DDB_TABLE: !Ref DynamodbTableArn
```
farski, line 110: Would also prefer a more descriptive environment variable name here

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+          DOVETAIL_HOST: !Ref DovetailHost
```
farski, line 111: Is this the Dovetail _router_ host? If so, maybe the name could reflect that.

farski (changes requested):
Added some thoughts

stacks/serverless/dovetail-cdn-arranger.yml
```diff
@@ -0,0 +1,133 @@
...
+  EnvironmentType:
+    Type: String
+  DovetailHost:
+    Type: AWS::SSM::Parameter::Value<String>
+  DovetailToken:
+    Type: AWS::SSM::Parameter::Value<String>
+  DynamodbTableArn:
+    Type: AWS::SSM::Parameter::Value<String>
+Resources:
+  DovetailCdnArrangerWorkspaceBucket:
```
farski, line 20: Because of how S3 bucket names are derived, this will almost certainly be truncated to something like `………-DovetailC`. It's useful to front load the resource logical name in this case with the most descriptive part, to make it easier to identify and find the bucket, even though it does break with resource naming conventions.

farski (changes requested):

stacks/serverless/dovetail-cdn-arranger.yml
cavis: +1 ... I didn't look into tag-specific expirations because I'm 90% sure I don't care anymore, and expiring everything in this bucket after a certain time is fine.

That said - the lambda still tags everything it PUTs, so we'd have the option of expiring sooner than the "global" expiration.

stacks/serverless/dovetail-cdn-arranger.yml
cavis: Bah- I think Atom+Prettier added all those.  Or at least changed them from double to single.  Will see if there's some other linting command that would remove.  Or i'll just remove them.

stacks/serverless/dovetail-cdn-arranger.yml
cavis: It's actually either (they both have the API).  So purposefully left a bit ambiguous.

stacks/serverless/dovetail-cdn-arranger.yml
cavis: Yeah, +1 to having a better way to version these.  But for now, I'm fine with ParameterStore translating to ENVs, just to get this out there.  These will hardly ever change, and in the edge stitch-lambda, I actually have them hardcoded in github.

stacks/serverless/dovetail-cdn-arranger.yml
cavis: Okay, so I'll rename the ENV to be more descriptive, change the ParameterStore param to be a table _name_ (not arn), and do some wildcard-ing in the IAM role (i think?).

And yes, this would ideally be a multi-region thing.  And probably the data account, though that's introducing some complexity with role-switching for all the lambdas using it.  I guess we'll get this one out there, then re-appraise how cross account DDB access works.

cavis:
Okay, I think I've resolved all the things.  And the project-tags for this are still a TODO (to split up the dovetail tagging monolith).

.prettierrc.yml
```diff
@@ -13,3 +13,6 @@ overrides:
     options:
       tabWidth: 4
       useTabs: true
+  - files: '*.yml'
```
farski, line 16: If these are still single quotes something's not right

farski (approved):

PR #478, opened 2021-02-19
Lint and style checks in CI

### LINKED ISSUE #104: Coding standards/linting
Mainly for JS and python.
If yaml (or even CFN templates) can be linted maybe that, too.

### LINKED ISSUE #480: Resolve all YAML issues from Prettier

### LINKED ISSUE #482: Add YAML files to Prettier checks in CI
Blocked by #480

farski:
Adds the following code checks and linting:

- Black for Python formatting
- flake8 for Python style checking
- TypeScript for JavaScript static analysis
- ESLint for JavaScript style
- Prettier for JavaScript and YAML formatting

All of the code changes in this PR are just resolving style errors.

Closes #104, Closes #480, Closes #482


.prettierrc.yml
```diff
@@ -16,3 +16,4 @@ overrides:
   - files: "*.yml"
     options:
       singleQuote: false
+      printWidth: 250
```
cavis, line 19: What does prettier even do if you go over the width limit?  (Maybe nothing... just error).

CONTRIBUTING.md
```diff
@@ -35,10 +35,34 @@ There is also an atom plugin https://atom.io/packages/atom-cfn-lint.
...
+1. When function nesting does not support short form, such as using `Fn::ImportValue` with `Fn::Join`.
+
+2. When it would lead to the function being the only thing on the line:
+
+```yaml
+# Undesirable
+Pets:
+  - Animal: Dog
+    Name: Scraps
+  - !If
```
cavis, line 51: Hrmm... but why can't you do this?  Is it just not allowed by CFN?

cdn/dovetail-cdn/dovetail-stitch-lambda-filters.yml
```diff
@@ -200,27 +192,20 @@ Resources:
       AlarmName: !Sub "[DovetailStitchLambda][Throttles] ${EnvironmentType} > 0"
       AlarmActions:
         - Fn::If:
-          - IsProduction
-          - Fn::ImportValue:
-              !Sub "${InfrastructureNotificationsStackName}-OpsErrorMessagesSnsTopicArn"
-          - Fn::ImportValue:
-              !Sub "${InfrastructureNotificationsStackName}-OpsWarnMessagesSnsTopicArn"
+            - IsProduction
```
cavis, line 195: Hrm... these seem too-indented now?  I would have thought prettier would remove 2 spaces.

cdn/dovetail-cdn/dovetail-stitch-lambda-filters.yml
```diff
@@ -240,27 +225,20 @@ Resources:
...
       OKActions:
         - Fn::If:
-          - IsProduction
-          - Fn::ImportValue:
-              !Sub "${InfrastructureNotificationsStackName}-OpsErrorMessagesSnsTopicArn"
-          - Fn::ImportValue:
-              !Sub "${InfrastructureNotificationsStackName}-OpsWarnMessagesSnsTopicArn"
-      AlarmDescription:
-        Logged errors on dovetail stitch lambda exceeded 0
+            - IsProduction
```
cavis, line 238: Yeah, a bunch of these indented 4 spaces from the `- Fn::If:`

cdn/single-origin.yml
```diff
@@ -347,10 +347,10 @@ Parameters:
       when the origin is an S3 bucket. Turning this on will not update any S3
       bucket policies, but such a policy is required to allow the origin access
       identity to function properly.
-    Default: 'No'
+    Default: "No"
     AllowedValues:
-      - 'Yes'
-      - 'No'
+      - "Yes"
+      - "No"
```
cavis, line 353: Seems like prettier will change single -> double quotes, but won't go as far as removing unnecessary quotes?  So I guess that'll just be a stylistic thing for whomever is writing the template.

package.json
```diff
@@ -11,20 +11,20 @@
...
     "eslint-config-airbnb-base": "^14.1.0",
-    "eslint-config-prettier": "^6.11.0",
+    "eslint-config-prettier": "^7.2.0",
     "eslint-plugin-import": "^2.20.2",
     "eslint-plugin-jest": "^24.1.0",
-    "eslint-plugin-prettier": "^3.1.3",
+    "eslint-plugin-prettier": "^3.3.1",
     "jest": "^26.0.1",
-    "prettier": "^2.0.5",
+    "prettier": "^2.2.1",
```
cavis, line 25: Is there a reason the lockfile in this project is ignored?  (Also ... npm vs yarn ... but we won't get into that!)

I think it's pretty heavily suggested that you lock linters to an exact version, to prevent churn between different users.

requirements.txt
```diff
@@ -0,0 +1,3 @@
+cfn-lint>=0.37.0
+black>=20.8b1
+flake8>=3.8.4
```
cavis, line 3: Ahhhh python.  I have the same question/comment here - should these be locked to an exact version?

stacks/ecs.yml
```diff
@@ -463,7 +462,7 @@ Resources:
               command: "(crontab -uroot -l 2> /dev/null; echo '* * * * * /root/cwdiskusage.sh') | crontab -u root -"
             03_fstrim_reclaim_disk:
               # http://docs.aws.amazon.com/AmazonECS/latest/developerguide/CannotCreateContainerError.html
-              command: "(crontab -uroot -l 2> /dev/null; echo '*/5 * * * * docker ps -qa | xargs docker inspect --format=\"{{ .State.Pid }}\" | xargs -IZ /sbin/fstrim /proc/Z/root/') | crontab -u root -"
+              command: '(crontab -uroot -l 2> /dev/null; echo ''*/5 * * * * docker ps -qa | xargs docker inspect --format="{{ .State.Pid }}" | xargs -IZ /sbin/fstrim /proc/Z/root/'') | crontab -u root -'
```
cavis, line 465: Oh hey, a situation where prettier prefers single quotes?

cavis (changes requested):
Huzzah for linting!  Couple questions and one suggestion on version locking.

I'm also wondering - what are the risks with deploying this?  The main 2 I can think of are:

1. Something in prod has drifted from Infrastructure, and touching these files resets it back to the canonical version.
2. Touching one of these stacks makes CFN _think_ it's changed even though it hasn't.  And it's one of those problematic-to-update stacks (like the VPC subnets).

I guess 2 is pretty unlikely.  And 1 is something i'd probably rather proactively find out about.

.prettierrc.yml
farski: In most cases it would find somewhere to break the line up into multiple lines. If it can't, the behavior is language dependent. For example, I've seen in JS files that it's ok with very long strings and comments, but in Python it doesn't seem to make that exception.

requirements.txt
farski: These are using a `>=-`, which I think is like `^` in npm. But I don't really know anything about requirements files, this is just how I've always seen it done.

cdn/single-origin.yml
farski: In this case these quotes aren't unnecessary. `Yes` and `No` are different than `"Yes"` and `"No"` in YAML, since the former are the same as `true`, `on`, `y`, and all the other boolean literal values that exist in YAML.

CONTRIBUTING.md
farski: It is allowed I just think it is harder to read. I find that having some actual value, like the string `Fn::If` on a line makes a lot more sense than a YAML tag which doesn't _really_ exist

stacks/ecs.yml
farski: Yeah, not sure why escaped single quotes is better than escaped double, but apparently it is.

package.json
farski: In the past there's been no reason to lock any of the packages in there, this does seem like a good time to start though. I'll add it.

cdn/dovetail-cdn/dovetail-stitch-lambda-filters.yml
farski: They are being indented relative to the `Fn::If:` key, not the `-` array item, which is actually the way I prefer it, and one of the reasons I use the full function name in these cases over the tag.

farski:
@cavis If No. 1 happens it's a price we'd need to pay at some point, and is our punishment for introducing drift. We can't get into situations where we're scared to deploy things.

But for both questions, none of the changes here should affect the resolved templates, meaning if they were converted to whitespace-compacted JSON they should stay the same. As long as that's the case, CloudFormation won't find anything it thinks it needs to do and won't make any API calls. If you see any changes that you're concerned it might construe as an actually resource change please flag them.

cavis (approved):
👍

PR #488, opened 2021-03-09
Cleanup prxu.org DNS zone template

farski:


dns/prxu.org-hosted_zone.yml
```diff
@@ -104,47 +99,26 @@ Resources:
...
       HostedZoneId: !Ref HostedZone
       RecordSets:
-        - Type: A
-          Name: !Sub userupload.${Domain}
-          AliasTarget:
-            DNSName: dsn15m3yob5tf.cloudfront.net.
-            # Global CloudFront hosted zone ID
-            HostedZoneId: Z2FDTNDATAQYW2
-        - Type: AAAA
-          Name: !Sub userupload.${Domain}
```
cavis, line 143: userupload?  Was this a `prx-up` thing?

cavis (approved):
This should be fine.

Just double-checked Highwinds - and `cdn-dovetail` gets maybe < 30 requests per month.  `ad-files` is getting 0.

dns/prxu.org-hosted_zone.yml
farski: It was a CloudFront that wasn't connected to anything

PR #490, opened 2021-03-11
Create CMS memcached in CMS stack

farski:
Moves CMS into its own template, rather than using the shared web-app template (necessary for IAM role changes as well).

Creates the CMS memcached cluster along side the other CMS app resources. Currently the cluster is created, but dormant, in production.

### DIFF
--- a/stacks/container-apps/cms.prx.org.yml
+++ b/stacks/container-apps/cms.prx.org.yml
@@ -344,6 +344,11 @@ Resources:
               Value: !Ref SecretsVersion
             - Name: AWS_DEFAULT_REGION
               Value: !Ref AWS::Region
+            - Fn::If:
+                - CreateProductionResources
+                - !Ref AWS::NoValue
+                - Name: MEMCACHE_SERVERS
+                  Value: !GetAtt CmsMemcacheCluster.ConfigurationEndpoint.Address
           Essential: true
           Image: !Sub ${AWS::AccountId}.dkr.ecr.${EcrRegion}.amazonaws.com/${AppName}.prx.org:${EcrImageTag}
           LogConfiguration:
--- a/stacks/container-apps/cms.prx.org.yml
+++ b/stacks/container-apps/cms.prx.org.yml
@@ -1,0 +1,392 @@
+# stacks/container-apps/web-application.yml
+AWSTemplateFormatVersion: "2010-09-09"
+Description: >
+  Docker web application with optional worker, using the shared platform ALB
+Conditions:
+  CreateWorkerResources: !Equals [!Ref CreateWorker, true]
+  CreateProductionResources: !Equals [!Ref EnvironmentType, Production]
+  HasMemoryReservation: !Not [!Equals [!Ref ContainerMemoryReservation, ""]]
+  HasAltName: !Not [!Equals [!Ref AltName, ""]]
+  HasExplicitHostHeader: !Not [!Equals [!Ref ExplicitHostHeader, ""]]
+Parameters:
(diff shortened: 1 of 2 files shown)


stacks/container-apps/cms.prx.org.yml
```diff
@@ -252,6 +252,11 @@ Resources:
               Value: !Ref SecretsVersion
             - Name: AWS_DEFAULT_REGION
               Value: !Ref AWS::Region
+            - Fn::If:
+                - CreateProductionResources
```
cavis, line 256: In any case - aws-secrets is probably going to overwrite your ENV-of-the-same-name here.  But +1 safety!

stacks/container-apps/cms.prx.org.yml
```diff
@@ -379,6 +384,58 @@ Resources:
...
+      SubnetIds:
+        - !Ref VPCSubnet1
+        - !Ref VPCSubnet2
+  CmsMemcacheSecurityGroup:
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      VpcId: !Ref VPC
+      GroupDescription: !Sub CMS ${EnvironmentType} memcached security group
+      SecurityGroupIngress:
+        - CidrIp: 0.0.0.0/0
```
cavis, line 406: Open to the world?  Or I guess just the VPC.

stacks/container-apps/cms.prx.org.yml
```diff
@@ -379,6 +384,58 @@ Resources:
...
+          FromPort: 11211
+          ToPort: 11211
+        - CidrIpv6: ::/0
+          IpProtocol: tcp
+          FromPort: 11211
+          ToPort: 11211
+  CmsMemcacheCluster:
+    Type: AWS::ElastiCache::CacheCluster
+    Properties:
+      AutoMinorVersionUpgrade: true
```
cavis, line 417: I usually leave this false, for apps that are may not be able to handle downtime.  But maybe CMS can?  What's the current setting?

stacks/container-apps/cms.prx.org.yml
```diff
@@ -0,0 +1,459 @@
+# stacks/container-apps/cms.prx.org.yml
+AWSTemplateFormatVersion: "2010-09-09"
+Description: >
+  Docker web application with optional worker, using the shared platform ALB
```
cavis, line 4: Do you planning to change this, and other copied-from-shared-stack things that don't need to be so specific?

Or are you just going for minimal changes after copy? I see you did leave all the configs in the parent stack, rather than moving them into here.

(And going forward, will we tend to keep configs in the lower/root stacks, or push them up to the specific app?)

cavis (approved):
Seems like it should work!  Couple questions...

stacks/container-apps/cms.prx.org.yml
farski: Copied what was in the memcache template that the existing cluster was launched from. If there's a more locked-down way to do this, especially now that it's in the same VPC, I'd be up for that, I'm just bad at VPCs.

stacks/container-apps/cms.prx.org.yml
farski: Even in prod, the Cfn template has this set to `true`. I can't really find what that maps to in console, so…yes?

stacks/container-apps/cms.prx.org.yml
farski: Leaving as is for me, will clean up the unnecessary parameters down the road.

PR #495, opened 2021-03-15
Build Castle Redis in CD stack

farski:


stacks/castle.prx.org.yml
```diff
@@ -225,6 +233,59 @@ Resources:
...
+        - !Ref VPCSubnet3
+  CastleRedisSecurityGroup:
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      VpcId: !Ref VPC
+      GroupDescription: !Sub Castle ${EnvironmentType} Redis security group
+      SecurityGroupIngress:
+        - IpProtocol: tcp
+          FromPort: 6379
+          SourceSecurityGroupId: !Ref VPCSecurityGroup
```
farski, line 254: I think this will work? May want to do it for CMS memcache this way too

cavis (approved):
Some of those CFN ElastiCache param names are unintuitive!  But looks like it might work to me.  +1!

PR #497, opened 2021-03-18
Consolidate multiple Exchange templates

farski:
This consolidates the primary, Say When, and worker auto scaling templates into the (newish) Exchange-specific template (which was recently refactored out of the shared web-app template for other reasons).

This will recreate all resources that were in the Say When and auto scaling templates from scratch.


stacks/container-apps/exchange.prx.org.yml
```diff
@@ -59,6 +63,7 @@ Parameters:
     Type: String
   CreateWorker:
```
cavis, line 64: Seems like you could just remove the `CreateWorker` param and all the `CreateWorkerResources` conditionals.  Since we always create the worker here.

stacks/container-apps/exchange.prx.org.yml
```diff
@@ -102,7 +108,17 @@ Parameters:
     Default: web
   HealthCheckPath:
     Type: String
-    Default: /api/v1
+    Default: /ping
```
cavis, line 111: Not sure if this is the time to address it ... but this stack feels kind of strange with all the internal-only params that are only looked up in 1 place.

stacks/container-apps/root.yml
```diff
@@ -173,69 +175,13 @@ Resources:
...
         EcrImageTag: !Ref ExchangeEcrImageTag
         SecretsVersion: !Ref ExchangeSecretsVersion
-        AppName: "exchange"
-        AltName: "api"
-        CreateWorker: "true"
-        DesiredWebsProduction: "5"
-        DesiredWorkersProduction: "6"
-        ErrorAlarmThreshold: "5"
-        ContainerPort: "3000"
-        HealthCheckPath: "/ping"
```
cavis, line 183: Ahh, I see you made these defaults.  Which is still a bit weird.  But hey, this is mostly a copy/paste PR.

cavis (changes requested):
The `CreateWorkerResources` seems kind of pointless?  And I'm curious for your thoughts on cleaning up the other redundant params someday.  Or maybe we just don't care.  It's not cool to care about extra params.

stacks/container-apps/root.yml
farski: The `AppName` param showed up in 29 places, which felt like too many to deal with right now, but FINE I'll do it

farski:
@cavis I think I got everything that didn't really make sense as a parameter anymore

cavis (approved):
👍  looks good to me.  The only thing that I'd think could mess up the deploy would be an ALB listener rule conflict, and they appear to keep the same names/priorities.

farski:
I don't think Cfn will even see these as changes to the listener rules, so hopefully the deploys don't touch them.

PR #500, opened 2021-03-23
Remove ECS service role

farski:
This removes all traces of the ECS service role, which has been replaced by an ECS service-link role that AWS manages, and was serving no real purpose.


cavis (approved):
Should be fine?  But I guess the risk is still that we've manually granted the `ECSServiceIAMRole` access to something, and removing it from the Service breaks something.

I think that's a pretty low risk.  But may want to deploy it when other things are quiet, and we can check that things still work.

farski:
These services are all being rewritten, and this change is reflected in the new versions. Closing this because it's not worth dealing with updating all the existing production services.

PR #503, opened 2021-04-27
Add Play proxy Lambda function to legacy stack

farski:
I have done very rudimentary testing on this, but it's fairly straight forward.

This only creates the `/proxy*` listener rule in staging, so it can safely be deployed to prod without switching to the new proxy.

`https://play.dualstack.infrast-alb-zh49czohymrr-1878189151.us-east-1.elb.amazonaws.com.stag.prx.tech/proxy2?url=http://feed.songexploder.net/SongExploder` is an example of the proxy running in the new cluster, **not** behind CloudFront like it will be in reality.

### DIFF
--- a/stacks/container-apps/play.prx.org.yml
+++ b/stacks/container-apps/play.prx.org.yml
@@ -9,6 +9,7 @@ Mappings:
       name: play-express
 Conditions:
   CreateProductionResources: !Equals [!Ref EnvironmentType, Production]
+  CreateProductionResources: !Equals [!Ref EnvironmentType, Staging]
 Parameters:
   # VPC ########################################################################
   VPC:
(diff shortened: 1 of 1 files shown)


stacks/container-apps/play.prx.org.yml
```diff
@@ -93,7 +95,7 @@ Resources:
           Values:
             - play.*
       ListenerArn: !Ref PlatformALBHTTPSListenerArn
-      Priority: !Join ["", [!Ref PlatformALBListenerPriorityPrefix, "00"]]
+      Priority: !Join ["", [!Ref PlatformALBListenerPriorityPrefix, "30"]]
```
cavis, line 98: Why change this?

stacks/container-apps/play.prx.org.yml
```diff
@@ -201,6 +203,159 @@ Resources:
...
+              path: `${q.pathname || ""}${q.search || ""}`,
+              method: "GET",
+              headers: {
+                Accept: ACCEPTED_TYPES.join(","),
+                "User-Agent": PROXY_UA,
+              },
+            };
+
+            const h = q.protocol === "https:" ? https : http;
+            const req = h.request(options, (res) => {
```
cavis, line 279: I don't think it matters, but the Play proxy currently uses the `request` library.  Which would have followed any redirects.  But in practice ... I can't imagine anyone is (or should be) using the player with a redirect url.

cavis: Oh wait, nevermind!  You do follow redirects below.

stacks/container-apps/play.prx.org.yml
```diff
@@ -201,6 +203,159 @@ Resources:
...
+
+          return {
+            statusCode: 200,
+            statusDescription: "200 OK",
+            isBase64Encoded: false,
+            headers: {
+              "Content-Type": "application/rss+xml",
+              "Cache-Control": "public, max-age=90",
+            },
+            body: res,
```
cavis, line 335: Isn't there a limit on response sizes for API Gateway?

cavis (commented):
Seems like it might work!  Just a couple minor questions...

stacks/container-apps/play.prx.org.yml
farski: If the host-only rule stays at a lower priority it will continue to catch all traffic

stacks/container-apps/play.prx.org.yml
farski: I'm not sure, but there's not API Gateway here

farski:
Didn't work out.

PR #505, opened 2021-05-24
CDN origin request lambda

cavis:
Initial setup for the to-be-edge-lambda'd "Dovetail CDN Origin Request".


farski (approved):
Seems good. I had hoped that the timing of the new multi-region refactored stacks and this would have aligned better, and we would only have had to do this work the new way, but I think we do have to manage this in both spots for now. I can take care of the new copy.

PR #509, opened 2021-06-17
Relax the success action redirect requirement

svevang:
Relaxes the redirect requirement for XHR style request ergonomics.


farski:
Why is this being relaxed? Or I guess why is this code changing at all? I thought we were replacing this app with native functionality in Augury.

svevang:
Looking at the evaporate v4 signing stuff. A signing endpoint is required (or at least very strongly suggested).  But it doesn't seem necessary to port this over to ruby? It looks like we can cleanly split the request signing out into this lambda (keep it here) and adopt an in-app uploader in augury.

The proximal answer is that the redirect flow doesn't really work that well with XHR requests, and we can get the same info with the `success_action_code` param.

PR #515, opened 2021-07-06
Consolidate the metrics due to Augury Agent forecast changes

svevang:
Merge this after https://github.com/PRX/augury.prx.org/pull/579 lands

This PR consolidates the metrics, reflecting the single invocation of the partitions forecast R script.


cavis (approved):
👍 seems right to me!  May need to update a couple of AWS "augury forecast" dashboards after this.  But they're already a bit stale.

PR #517, opened 2021-07-07
Make sure alarm/metrics namespaces don't collide

svevang:
Took another pass related to https://github.com/PRX/augury.prx.org/issues/527 on clearing up collisions in the metrics / filters namespaces.

### DIFF
--- a/stacks/apps/augury.yml
+++ b/stacks/apps/augury.yml
@@ -364,23 +364,6 @@ Resources:
         - MetricName: !Sub InventoryUpdateJobElapsed${EnvironmentType}
           MetricNamespace: PRX/Augury
           MetricValue: $.elapsed
-  AuguryActualsAreBehindAlarm:
-    Type: AWS::CloudWatch::Alarm
-    Properties:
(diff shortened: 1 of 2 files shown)


farski (commented):
I think this works, and don't want to block the merge, but I would probably be inclined to change the `MetricName` of the resources in the new cluster (`/apps/augury.yml`) instead of these ones from the old cluster. These are still the live, production resources, it feels a little safer to mess with the resources that are just sitting there waiting, but aren't actually doing anything yet. Also if these metrics in the PR _don't_ get renamed, we maintain the continuity of the existing data. I'm a 3/10 on this, because ultimately I don't think it matters very much, I've just been trying to leave the existing resource untouched unless it's unavoidable, and do work on the new stuff while it's still germinating.

If we knew the new resources were going to be going live very soon, I'd say it's safe to just treat them that way now, but I think we're probably still a ways off from these actually being called up for service.

svevang:
Changed the name of the metric, but now just in the new cluster defs!

farski (approved):

PR #520, opened 2021-07-13
Staging feeder on aurora

cavis:
DO NOT MERGE YET.

For PRX/internal#715


stacks/container-apps/feeder.prx.org.yml
```diff
@@ -78,6 +79,10 @@ Parameters:
   HealthCheckPath:
     Type: String
     Default: /api/v1
+  SharedRdsPostgresqlEndpoint:
```
farski, line 82: Is this needed?

cavis: I wired it through, thinking we'd switch Prod over to using that param now.  And use the `Fn::If` to toggle prod to RDS and stag to Aurora.

But maybe not?

farski: No that makes sense.

farski (approved):

PR #522, opened 2021-07-14
Reprioritize ALB listener rules

farski:
This is based on recently observed data


cavis (approved):
👍 don't see any conflicting numbers, so looks good to me.

PR #529, opened 2021-08-09
Alarm on logged warnings/errors

cavis:


farski (changes requested):
The name for the alarms should match the `prx:dev:application` tag, so:
- `[Dovetail-CDN Arranger]`
- `[Dovetail-CDN Origin Request]`

cdn/dovetail3-cdn.yml
```diff
@@ -358,7 +358,7 @@ Resources:
       Namespace: AWS/CloudFront
       Period: 60
       Statistic: Average
-      Threshold: 0.5
```
cavis, line 361: Seemed a bit weird to look at these fractional percentages.  These will probably need updating as we get more traffic moved over anyways.

farski (approved):

PR #531, opened 2021-08-23
Permanently delete non-current object versions

farski:
The mediajoint, networks, and feeds buckets previously did not have versioning enabled. In order to support replication, version had to be enabled. This will create a lifecycle rule that expires (i.e., permanently deletes) non-current versions of all objects in the bucket, which will prevent any unnecessary buildup of old object versions.


cavis (approved):
👍 works for me!

PR #532, opened 2021-08-31
Staging mysql aurora

cavis:
For PRX/internal#723.

May need to add some params to the zipfile first.


farski:
Should this PR to apps/exchange and apps/cms too?

cavis:
Ah, yes!  I was just trying to un-confuse myself about the order these ENV changes get deployed in.

PR #534, opened 2021-09-02
Use stack params for db host/port

cavis:
Consistently use stack params for setting the db hosts/ports, instead of AWS secrets.  Note that prod is still pointed at non-Aurora at this point.

Deploying this:

- [x] Ensure the prod zipfile contains the shared mysql/postgres host variables.
- [ ] Deploy this (aws-secrets hosts/ports in prod will continue to override this)
- [ ] Update every app's aws-secrets, removing the host/port ENVs

### DIFF
--- a/stacks/apps/augury.yml
+++ b/stacks/apps/augury.yml
@@ -41,7 +41,6 @@ Parameters:
   SharedAuroraPostgresqlEndpoint: { Type: String }
 
 Conditions:
-  IsStaging: !Equals [!Ref EnvironmentType, Staging]
   IsProduction: !Equals [!Ref EnvironmentType, Production]
   IsPrimaryRegion: !Equals [!Ref RegionMode, Primary]
   HasRdsEndpoint: !Not [!Equals [!Ref SharedRdsPostgresqlEndpoint, ""]]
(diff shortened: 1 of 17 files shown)


stacks/apps/augury.yml
```diff
@@ -41,7 +41,6 @@ Parameters:
   SharedAuroraPostgresqlEndpoint: { Type: String }
 
 Conditions:
-  IsStaging: !Equals [!Ref EnvironmentType, Staging]
   IsProduction: !Equals [!Ref EnvironmentType, Production]
   IsPrimaryRegion: !Equals [!Ref RegionMode, Primary]
   HasRdsEndpoint: !Not [!Equals [!Ref SharedRdsPostgresqlEndpoint, ""]]
```
cavis, line 46: @farski should i make any attempt to clean up these guards?  I see a mix of `Condition HasRdsEndpoint not used` and `Condition HasAuroraEndpoint not used` when I lint these new app stacks.

farski: I wouldn't worry about `HasRdsEndpoint` much, that will be going away once everything is off RDS in staging and prod.

`HasAuroraEndpoint` would be worth double checking at some point. I think anything that's using either Aurora endpoint value should be behind a `HasAuroraEndpoint` condition.

farski (changes requested):
I think the port values should be moved to stack parameters as well, so that entire configuration is coming from the same place. But I'm okay with punting on that until we're down to two databases.

farski (approved):

PR #555, opened 2021-11-05
Refactor CMS IAM access into task role

farski:
Refactors the policies from IAM user `prx.org-backend` into the ECS task role.

### DIFF
--- a/stacks/apps/cms.yml
+++ b/stacks/apps/cms.yml
@@ -170,13 +170,11 @@ Resources:
           PolicyName: Announce
         - PolicyDocument:
             Statement:
-              - Action: s3:ListAllMyBuckets
-                Effect: Allow
-                Resource: "*"
               - Action:
                   - s3:GetObject
                   - s3:GetObjectVersion
                   - s3:ListBucket
+                  - s3:ListAllMyBuckets
                 Effect: Allow
                 Resource: "*"
             Version: "2012-10-17"
--- a/stacks/apps/cms.yml
+++ b/stacks/apps/cms.yml
@@ -193,8 +193,8 @@ Resources:
                   - s3:RestoreObject
                 Effect: Allow
                 Resource: # Turn these into root stack paramters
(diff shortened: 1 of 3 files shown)


stacks/apps/cms.yml
```diff
@@ -137,6 +138,65 @@ Resources:
...
+        - PolicyDocument:
+            Statement:
+              - Action: sqs:*
+                Effect: Allow
+                Resource:
+                  - !GetAtt AudioCallbackQueue.Arn
+                  - !GetAtt ImageCallbackQueue.Arn
+                  - !GetAtt PodcastImportQueue.Arn
+                  - !GetAtt SearchIndexerQueue.Arn
+                  - !GetAtt DefaultJobQueue.Arn
```
cavis, line 158: Does this cover dead letter queues?  Or I guess ... that's built into SQS, and CMS would never pick up a message from one.  Nevermind!

stacks/apps/cms.yml
```diff
@@ -137,6 +138,65 @@ Resources:
...
+                  - s3:AbortMultipartUpload
+                  - s3:DeleteObject
+                  - s3:Get*
+                  - s3:List*
+                  - s3:Put*
+                  - s3:RestoreObject
+                Effect: Allow
+                Resource: # Turn these into root stack paramters
+                  - !If [IsProduction, arn:aws:s3:::prx-feed/*, arn:aws:s3:::staging-prx-feed/*]
+                  - !If [IsProduction, arn:aws:s3:::prx-feed, arn:aws:s3:::staging-prx-feed]
```
cavis, line 197: CMS needs write access to the feed bucket?  I thought it might read from other buckets ... but writes were always to the mediajoint bucket.

farski: This policy applied to both `prx-feed` and `prx-up` in the user, so that could have something to do with it. I copied the policies as they were, but would be glad for them to get tightened up if possible.

PR #565, opened 2021-11-22
Modernize Dovetail CDN tags

farski:


cavis (approved):
👍  Fine by me!  May not be worth redeploying the old CDN, as it's not long for this world.

Hopefully doesn't tear anything down!

PR #572, opened 2021-12-07
Enable Docker init for Metrics

farski:


cavis (approved):
👍 Sure!  Is the plan to enable this on our other apps too?

farski:
Yeah, if the changes I made for Metrics on both sides are correct, and functionally identical to how they are now, I was going to do the rest throughout the week

PR #585, opened 2022-02-17
ECS draining

cavis:
Bring back the [ECS task draining](https://github.com/PRX/Infrastructure/blob/bcdbefed9ad2cddf5bc8d1112b254918789e06c1/stacks/ecs.yml#L653) lambda/SQS.  Attempts to convert to `AWS::Serverless::Function` - we'll see if that works!

Just staging for now.

### DIFF
--- a/stacks/shared-ecs/asg.yml
+++ b/stacks/shared-ecs/asg.yml
@@ -103,7 +103,7 @@ Resources:
   # this lambda handles draining ECS tasks from instances in this ASG before they
   # terminate. Should keep an eye on the github issue for less-DIY support of this
   # functionality: https://github.com/aws/containers-roadmap/issues/256
-  ECSDrainSNSTopic:
+  EcsDrainSnsTopic:
     Condition: IsStaging
     Type: AWS::SNS::Topic
     Properties:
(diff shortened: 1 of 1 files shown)


stacks/shared-ecs/asg.yml
```diff
@@ -99,6 +99,333 @@ Resources:
...
+                                "Response received from complete_lifecycle_action %s", response
+                            )
+                            logger.info("Completedlifecycle hook action")
+                        except Exception as e:
+                            print(str(e))
+      MemorySize: 256
+      Policies:
+        - Statement:
+            - Action:
+                - ec2:DescribeInstances
```
cavis, line 361: Don't know if it matters, but I dropped the [logging permissions](https://github.com/PRX/Infrastructure/blob/bcdbefed9ad2cddf5bc8d1112b254918789e06c1/stacks/ecs.yml#L695) here.

stacks/shared-ecs/asg.yml
```diff
@@ -99,6 +99,333 @@ Resources:
...
+                - ec2:DescribeInstanceAttribute
+                - ec2:DescribeInstanceStatus
+                - ec2:DescribeHosts
+                - ecs:DescribeContainerInstances
+                - ecs:DescribeTasks
+                - ecs:ListContainerInstances
+                - ecs:ListTasks
+                - sns:ListSubscriptions
+              Effect: Allow
+              Resource: "*"
```
cavis, line 371: Some of these seem a bit permissive?  At least they're read permissions.  May want to lock these down a bit though, once I prove this (still) works.

stacks/shared-ecs/asg.yml
```diff
@@ -99,6 +99,333 @@ Resources:
...
+          Version: "2012-10-17"
+        - Statement:
+            - Action:
+                - autoscaling:CompleteLifecycleAction
+              Effect: Allow
+              Resource: "*" # TODO: how to get an ASG ARN?
```
cavis, line 377: I guess there's no return value for the ASG ARN?

PR #586, opened 2022-02-21
TheCastle worker

kookster:
Added to `the-castle.yml` the def for a worker task that can run hourly.
Moved the EFS volume from the web service to the worker which needs it.
Updated parent stacks for additional param, and for having real ecr and secrets values for the new app stack

### DIFF
--- a/stacks/apps/the-castle.yml
+++ b/stacks/apps/the-castle.yml
@@ -369,7 +369,7 @@ Resources:
           EcsParameters:
             TaskCount: 1
             TaskDefinitionArn: !Ref WorkerTaskDefinition
-          Id: ThecastleWorkerRuleTarget
+          Id: TheCastleWorkerRuleTarget
           Input: !Sub | # containerOverrides comes from the ECS run-task API
             {
               "containerOverrides": [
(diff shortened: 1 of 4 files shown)


farski (approved):
I think that's looking pretty good

stacks/apps/the-castle.yml
```diff
@@ -11,6 +11,9 @@ Parameters:
   kWebContainerName:
     Type: String
     Default: thecastle-web
+  kWorkerContainerName:
+    Type: String
+    Default: thecastle-worker
   kWebApplicationPort:
     Type: Number
     Default: 8000
```
farski, line 19: This may need to be 3000

kookster: I think you're right

kookster: fixed

farski (commented):
L69 needs a real path to health check

farski (changes requested):
I would keep everything as `TheCastle`, except for the secrets version parameter in the root stack (and the root stack only)

farski (approved):
Possibly the web task needs more memory, but otherwise this 👍

PR #587, opened 2022-02-28
Add a 60 second lock timeout to the augury webs connection

svevang:
Related to: https://github.com/PRX/augury.prx.org/pull/869/files

### DIFF
--- a/stacks/apps/augury.yml
+++ b/stacks/apps/augury.yml
@@ -631,7 +631,7 @@ Resources:
               Value: "true"
             - Name: POSTGRES_POOL_SIZE
               Value: !Ref kPostgresPoolSize
-            - Name: POSTGRES_LOCK_TIMEOUT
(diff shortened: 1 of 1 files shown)


stacks/apps/augury.yml
```diff
@@ -631,6 +631,8 @@ Resources:
               Value: "true"
             - Name: POSTGRES_POOL_SIZE
               Value: !Ref kPostgresPoolSize
+            - Name: POSTGRES_STATEMENT_TIMEOUT
+              Value: 60000
```
farski, line 635: My guess is that this value isn't in seconds, so I'd maybe add a comment indication the scale or units

farski, line 635: My guess is that this value isn't in seconds, so I'd maybe add a comment indication the scale or units

farski (approved):

PR #589, opened 2022-03-01
Feat/bump augury webs

svevang:
Since most of the Augury activity is clustered around "bursts" late in the day UTC,  Let the auto scaler increment the number of web nodes to handle the "open in tabs and modify" behavior.

The flip side is that when more if the performance enhancements land on Augury we can tune this back down.


cavis (approved):
👍

PR #590, opened 2022-03-09
Bump the evaluation periods to silence EOD alarm

svevang:
Every day, in the forecasting quiescence of UTC evening, the forecast alarm goes off.  This PR adds another evaluation period, to make sure we are sampling slowly enough that a false positive is not triggered in the lull.


cavis (approved):
:+10:

PR #599, opened 2022-04-13
Jump server in every region

cavis:
Adds some regional DNS entries for the jump servers and dovetail-router:

```
jump.us-east-1.prx.tech
jump.staging.us-east-1.prx.tech

dovetail-router.us-east-1.prx.tech
dovetail-router.staging.us-east-1.prx.tech
```

Also makes the castle dependency on the jump servers optional.

### DIFF
--- a/stacks/apps/dovetail-router.yml
+++ b/stacks/apps/dovetail-router.yml
@@ -116,6 +116,7 @@ Resources:
       # https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy.html#routing-policy-latency
       # https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-latency-alias.html
       ServiceDomain: !Sub dovetail-router-alb-latency-group.${EnvironmentTypeAbbreviation}.prx.tech
+      VanityDomain: !If [IsProduction, !Sub "dovetail-router.${AWS::Region}.prx.tech", !Sub "dovetail-router.staging.${AWS::Region}.prx.tech"]
 
   Certificate:
     # This certificate does not include an entry for the service domain, since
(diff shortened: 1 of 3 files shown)


stacks/apps/dovetail-router.yml
```diff
@@ -172,6 +173,25 @@ Resources:
           SetIdentifier: !Ref AWS::StackName
           Type: A
 
+  VanityRecordSetGroup:
```
farski, line 176: Do you imagine these being used for anything other than a human accessing a specific region for testing?

cavis: I could see using them to get multi-region DTR off the ground.  Or even for monitoring/alarms.

farski: I would prefer to use the existing DNS entries for the regional Dovetail ALBs for those sorts of things, since they already exist, and there's no risk of them drifting or changing over time (or when they do it's explicitly because the underlying resource changed). It's also more consistent with how we handle the other ALBs. I'm fine with having some convenience/vanity domains to make devs' lives easier, but I don't want them to sneak into critical paths.

cavis: Yeah, fair enough.  So my first use would really be devtools - the various DTR debugging scripts, and `awssh` through the jump server stuff.

stacks/root.yml
```diff
@@ -948,7 +945,7 @@ Resources:
         KeyPairName: !Ref AsgKeyPairName
         RootStackName: !GetAtt Constants.RootStackName
         RootStackId: !GetAtt Constants.RootStackId
-        CastlePostgresClientSecurityGroupId: !GetAtt Apps200AStack.Outputs.CastlePostgresClientSecurityGroupId
+        CastlePostgresClientSecurityGroupId: !If [CreateApplicationStacks, !GetAtt Apps200AStack.Outputs.CastlePostgresClientSecurityGroupId, ""]
```
cavis, line 948: Will this work, or will CFN be angry that a `AWS::EC2::SecurityGroup::Id` parameter gets set to a blank string?
```

farski: Can't say for sure

farski (approved):

PR #603, opened 2022-04-19
New analytics DDB table

cavis:
Starting to move the analytics DDB from the data accounts into this stack.  Reasoning being: this "temporary cache of DTR redirect data" should be tied to a specific region/Dovetail-stack.

So this PR:

- Creates a new table directly in the `dovetail-analytics.yml` stack
- Sets up capacity scaling (hopefully) the same way as the data accounts
- Adds a temporary redundant `analytics-dynamodb-lambda` function, hooked up to the same kinesis stream as the real thing.
- New lambda will insert (hopefully) the same data into the new table.  But has no output subscription-filter.
- In > 24 hours (hopefully) our tables will reach parity.  And we can point the real function at the new table, and delete this temporary function.

### DIFF
--- a/stacks/apps/dovetail-analytics.yml
+++ b/stacks/apps/dovetail-analytics.yml
@@ -347,6 +347,74 @@ Resources:
         - { Key: prx:dev:application, Value: Analytics }
 
   # DynamoDB
+  AnalyticsDynamoDbTable:
+    Type: AWS::DynamoDB::Table
+    Properties:
+      AttributeDefinitions:
+        - AttributeName: id
+          AttributeType: S
+      BillingMode: PROVISIONED
+      KeySchema:
+        - AttributeName: id
+          KeyType: HASH
+      ProvisionedThroughput:
+        ReadCapacityUnits: !If [IsProduction, 100, 1]
+        WriteCapacityUnits: !If [IsProduction, 400, 1]
+      Tags:
+        - { Key: prx:meta:tagging-version, Value: "2021-04-07" }
+        - { Key: prx:cloudformation:stack-name, Value: !Ref AWS::StackName }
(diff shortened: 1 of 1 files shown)


stacks/apps/dovetail-analytics.yml
```diff
@@ -347,6 +347,276 @@ Resources:
...
+      AttributeDefinitions:
+        - AttributeName: id
+          AttributeType: S
+      BillingMode: PROVISIONED
+      KeySchema:
+        - AttributeName: id
+          KeyType: HASH
+      ProvisionedThroughput:
+        ReadCapacityUnits: !If [IsProduction, 100, 1]
+        WriteCapacityUnits: !If [IsProduction, 400, 1]
```
cavis, line 362: I'm not entirely sure how we got the RCU/WCU numbers for prod?  It's currently scaled all the way down to 1 on read capacity.  And all the way down to 215 (the minimum we specified) for write capacity.  Maybe trying to handle traffic spikes?

stacks/apps/dovetail-analytics.yml
```diff
@@ -347,6 +347,276 @@ Resources:
...
+        - { Key: prx:dev:family, Value: Dovetail }
+        - { Key: prx:dev:application, Value: Analytics }
+      TimeToLiveSpecification:
+        AttributeName: expiration
+        Enabled: true
+  AnalyticsDynamoDbTableWriteScalableTarget:
+    Type: AWS::ApplicationAutoScaling::ScalableTarget
+    Properties:
+      MaxCapacity: 1000
+      MinCapacity: !If [IsProduction, 215, 1]
```
cavis, line 379: Yeah, this is how it's currently setup.  But not sure it's right.  Why does ReadCapacity have a min of 1 in prod then?

PR #606, opened 2022-04-28
Set up alarms jobs and cache size

svevang:
Sets up Augury alarms for:
- Job Errors
- Allocation Cache Size

### DIFF
--- a/stacks/apps/augury.yml
+++ b/stacks/apps/augury.yml
@@ -75,6 +75,8 @@ Resources:
     Properties:
       ServiceToken: !Ref EchoServiceToken
       ActualsIngestJobInvocationCountMetricName: !Sub ActualsIngestJobInvocationCountNewCluster${EnvironmentType}
+      JobErrorsCountMetricName: !Sub JobErrorsCountNewCluster${EnvironmentType}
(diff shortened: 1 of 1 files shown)


stacks/apps/augury.yml
```diff
@@ -534,6 +556,34 @@ Resources:
...
+      EvaluationPeriods: 3
+      MetricName: !Sub Constants.TargetingRecordCacheSizeMetricsName
+      Namespace: !Ref kMetricFilterNamespace
+      Period: 300
+      Statistic: Maximum
+      Threshold: 1000
+      TreatMissingData: notBreaching
   AuguryActualsAreBehindAlarmTags:
     Type: Custom::CloudWatchAlarmTags
     Condition: IsProduction
```
farski, line 589: Also noting that these tags are conditional, but the alarm itself isn't. I know it's unrelated to this PR, but if you wanted to remove this condition that would be good.

farski (changes requested):

farski (changes requested):
I'm still not seeing the tags. Perhaps they didn't get included in the commit?

farski (approved):

PR #607, opened 2022-05-06
DDB global table

cavis:
Stack to deploy a DDB global table, in up to 3 regions.  The alarms for provisioned mode are still TBD.

Plan is to create a new "arrangements" table for dovetail using this, but in on-demand mode.


db/dynamodb-global-table.yml
```diff
@@ -0,0 +1,337 @@
...
+                  ScaleOutCooldown: !Ref AutoScalingCooldown
+                  TargetValue: !Ref ReadUtilizationScalingTarget
+            - !Ref AWS::NoValue
+          Region: !Ref AWS::Region
+          Tags:
+            - { Key: prx:meta:tagging-version, Value: "2021-04-07" }
+            - { Key: prx:cloudformation:stack-name, Value: !Ref AWS::StackName }
+            - { Key: prx:cloudformation:stack-id, Value: !Ref AWS::StackId }
+            - { Key: prx:ops:environment, Value: !Ref EnvironmentType }
+            - { Key: prx:dev:family, Value: !Ref FamilyTag }
```
farski, line 221: I don't imagine us using this template for much else, so probably fine for now, but the `family` tag should be optional.

db/dynamodb-global-table.yml
```diff
@@ -0,0 +1,337 @@
...
+  # Alarms
+  # TODO: provisioned billing might benefit from alarms - but they might have
+  # to exist in multiple regions to cover all these tables
+Outputs:
+  RoleArn:
+    Condition: HasForeignAccount
+    Description: The created DynamoDB table ARN
+    Value: !GetAtt CrossAccountAccessRole.Arn
+  TableArn:
+    Description: The created DynamoDB table ARN
```
farski, line 336: This would be the ARN for just _one_ of the tables that's created, right?

farski (approved):
lgtm

db/dynamodb-global-table.yml
cavis: Oh, yeah, I guess it is.  Maybe I'll change this output to `TableName` anyways ... that's what our lambdas take as an input.  The table name, region, and access role arn.

PR #609, opened 2022-05-09
Add access to s3 bucket for writing logs

kookster:
Maybe I did the permissions right?


farski (approved):
Looks pretty reasonable to me

PR #611, opened 2022-05-13
Add EC2 key pair

farski:
This creates the key pair, but does not use it anywhere yet

### DIFF
--- a/stacks/shared-ecs/asg.yml
+++ b/stacks/shared-ecs/asg.yml
@@ -436,6 +436,20 @@ Resources:
       IpProtocol: tcp
       ToPort: 65535
 
+  AsgInstanceKeyPair:
+    Type: AWS::EC2::KeyPair
+    Properties:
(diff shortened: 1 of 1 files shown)


stacks/shared-ecs/asg.yml
```diff
@@ -436,6 +436,20 @@ Resources:
       IpProtocol: tcp
       ToPort: 65535
 
+  AsgInstanceKeyPair:
```
cavis, line 439: I'm curious how you get the keypair.  Is it an output in CFN?  Are there security implications to it being visible there forever?

cavis (approved):
Works for me.

Would we create a different one for the jump-server then?  And potentially any other EC2 instances launched outside of the ASG.  Could also move this keypair up a level, to share it between them.

stacks/shared-ecs/asg.yml
farski: It puts the secret in Parameter Store when it's created. it's never available by CFN

PR #612, opened 2022-05-13
Permissions for multiple regions

cavis:


stacks/apps/dovetail-cdn-origin-request.yml
```diff
@@ -54,12 +54,16 @@ Resources:
         - Statement:
             - Action: lambda:InvokeFunction
               Effect: Allow
-              Resource: !Ref ArrangerFunctionArn # TODO multiple region arrangers
+              Resource:
+                - !Ref ArrangerFunctionArn
```
cavis, line 58: I opted to leave the explicit ARNs for this stack's us-east-1 arranger/bucket.  Just in case the name of something changes, at least we'd be able to invoke us-east-1.

farski: That makes sense, but I'd add a comment so we remember why it's there in three years

farski (approved):

PR #614, opened 2022-05-18
Add some ENVs to jump servers

cavis:
Add some envs to un-hardcode things in the `awssh` and `awstunnel` scripting.  Given a region and environment-type, they should be able to ssh you into the jump server and pull the configs you need.

I'd also like to get all `awssh` traffic going _through_ the jump server instead of having port 22 open on every ECS instance.  Someday...

### DIFF
--- a/stacks/jump-server.yml
+++ b/stacks/jump-server.yml
@@ -106,7 +106,6 @@ Resources:
                 [main]
                 stack=${AWS::StackId}
                 region=${AWS::Region}
-              mode: "000400"
               owner: root
               group: root
             /etc/cfn/hooks.d/cfn-auto-reloader.conf:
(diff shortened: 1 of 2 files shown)


stacks/jump-server.yml
```diff
@@ -87,6 +96,8 @@ Resources:
                   echo "Adding developer public keys to authorized_keys"
                   echo "${developer_keys}" >> /home/ec2-user/.ssh/authorized_keys
                 - developer_keys: !Join ["\n", !Ref AuthorizedKeys]
+            02_load_prx_env:
+              command: "printf \"\n# Load PRX environment variables\nif [ -f ~/.prxenv ]; then\n  . ~/.prxenv\nfi\n\" >> .bash_profile"
```
farski, line 100:
```yaml
command: |-
  printf…
```

makes this a little bit more readable without the escaped quotes

farski (approved):

PR #617, opened 2022-05-24
Switch Castle hostname to castle.prx.org

farski:


kookster (approved):
all 4 characters lgtm

PR #618, opened 2022-06-06
Create SSM VPC Endpoint

farski:
Adds a VPC Endpoint for SSM.

Also gives the shared ECS security group access to the endpoint.

Also gives explicit access to Augury slow worker and Dovetail Analytics Redis


cavis (approved):
👍

PR #621, opened 2022-06-14
Use only new hostname for The Count

farski:
- Updates shared hostname value to new domain, which is used in `COUNT_HOST` envar
- Removes `COUNT_HOST_NEW` envar


kookster (approved):
lgtm

PR #627, opened 2022-07-15
Run say_when as a cloudwatch event cron

cavis:


stacks/apps/exchange.yml
```diff
@@ -1147,7 +953,30 @@ Resources:
...
+    Condition: EnableWorkers # See README
+    Properties:
+      Description: !Sub Periodically runs ${EnvironmentType} SayWhen jobs
+      ScheduleExpression: rate(1 hour)
+      State: ENABLED
+      Targets:
+        - Arn: !Ref EcsClusterArn
+          EcsParameters:
+            TaskCount: 1
+            TaskDefinitionArn: !Ref WorkerTaskDefinition
```
cavis, line 967: Oh, I see.  Reuse the worker task def, but run a different command.

stacks/apps/exchange.yml
```diff
@@ -1391,6 +1220,7 @@ Resources:
         - { Key: prx:ops:environment, Value: !Ref EnvironmentType }
         - { Key: prx:dev:application, Value: Exchange }
       VisibilityTimeout: 5400 # 1.5 hour
+  # TODO: remove
```
cavis, line 1223: I guess this TODO is no longer valid either - I think I assumed we wouldn't need the queue at all anymore.  But if i understand this, the `process_jobs` will _add_ jobs to the say_when queue, to be processed by the other/real workers.  Right?

cavis (commented):
I think :+3:.  Though i'm the author of this PR, so github won't actually let me approve it.

stacks/apps/exchange.yml
kookster: yes! which is why we still need el queue de hablar donde

stacks/apps/exchange.yml
kookster: right, I don't think it has any other unique needs

kookster:
@farski can you review this as well?
I think it makes sense, and we didn't miss anything?

kookster:
I just merged in `main`, hopefully does something good for the build (and merging)

farski (approved):
Seems good

PR #628, opened 2022-07-18
Clean up DNS for The Count

farski:
- Removes legacy `count.prx.org` Contegix DNS record
- Replaces hard-coded `thecount.prx.org` value with the hostname parameter, which is now also `thecount.prx.org`


cavis (approved):
👍

PR #635, opened 2022-08-04
Remove some legacy Contegix DNS records

farski:
This removes DNS records where an HTTP request in the browser to the domain's root path did nothing useful.

For example, `newhome.prx.org` was removed because going to `newhome.prx.org` in a browser results in a maintenance page. Many of these domains resulted in maintenance pages. Some resulted in TLS or other connection errors. At least one led to a blank page.

None of these domains were tested on any other protocols or paths other than `/`. For example, `svn.prx.org` was not testing using Subversion.


kookster (approved):
these all look fine to me, none of these are live and public.
if we do need to connect to svn or something, can edit /etc/hosts

PR #643, opened 2022-09-12
Add redirect prefix to non-primary regions

cavis:
Leaving us-east-1 without a `REDIRECT_PREFIX` for now.  CloudFront knows that us-east-1 is the "default" region to send a realtime log to, when a request has no `/use1/` or `/usw2/` prefix/behavior.

### DIFF
--- a/stacks/apps/dovetail-router.yml
+++ b/stacks/apps/dovetail-router.yml
@@ -93,6 +93,7 @@ Parameters:
   AuguryHostname: { Type: String }
   FeederHostname: { Type: String }
   DovetailCdnHostname: { Type: String }
+  DovetailCdnRedirectPrefix: { Type: AWS::SSM::Parameter::Value<String> }
 
 Conditions:
   IsProduction: !Equals [!Ref EnvironmentType, Production]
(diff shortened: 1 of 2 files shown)


stacks/apps/dovetail-router.yml
```diff
@@ -795,6 +796,8 @@ Resources:
               Value: !Ref kApplicationPort
             - Name: REDIRECT_HOST
               Value: !Ref DovetailCdnHostname
+            - Name: REDIRECT_PREFIX
+              Value: !If [IsPrimaryRegion, !Ref "AWS::NoValue", !Ref DovetailCdnRedirectPrefix]
```
farski, line 800: Is it the primary region that has no prefix? Or us-east-1?

cavis: The primary region (which is us-east-1).  Which more importantly, is the "Primary Region" in the CDN template.

image

This may just be temporary.  I don't want to change too much with this deploy.  But once things are working, I may get rid of the `!If` here.

farski: Oh ok. We definitely shouldn't have anything in Spire that assumes PrimaryRegion always equals us-east-1. That's true now, but the moment where that becomes not true will be when we're dealing with some major issue, so not the time we want to be tracking down a weird behavior change.

cavis: I did already set the paramstore values for `/us-east-1/` too.  So once the `!If` is gone, it should change to the `use1` prefix.

farski: rgr

PR #646, opened 2022-09-26
Add logging and better scaling for DTR

cavis:


stacks/apps/dovetail-router.yml
```diff
@@ -674,8 +674,8 @@ Resources:
       Cluster: !Ref EcsClusterArn
       DeploymentConfiguration:
         MaximumPercent: 200
-        MinimumHealthyPercent: 50
-      DesiredCount: !If [IsProduction, 8, 2]
+        MinimumHealthyPercent: 100
+      DesiredCount: !If [IsProduction, 16, 2]
```
cavis, line 678: I've noticed 2 things during deploys:

1. ECS tends to immediately kill half the DTR tasks, while launching new ones. That might be fine for other apps, but it can (right now) cause the remaining DTR tasks to be overwhelmed.
2. I've been manually changing the min-task count in autoscaling to 16.  But deploying keeps setting it back to 8, which is not ideal.

stacks/apps/dovetail-router.yml
```diff
@@ -885,8 +887,8 @@ Resources:
   ScalableTarget:
     Type: AWS::ApplicationAutoScaling::ScalableTarget
     Properties:
-      MinCapacity: !If [IsProduction, 8, 2]
-      MaxCapacity: !If [IsProduction, 24, 4]
+      MinCapacity: !If [IsProduction, !If [IsPrimaryRegion, 16, 12], 2]
```
cavis, line 890: These are basically the rules I've been manually changing things to for awhile.

stacks/apps/dovetail-router.yml
```diff
@@ -947,10 +949,9 @@ Resources:
         const SERVICE_NAME = process.env.ECS_SERVICE_NAME;
         const SCALE_TO = {
           '815/feed-rss.xml': 4,
-          '284/feed-rss.xml': 9,
-          '70/feed-rss.xml': 24,
+          '70/feed-rss.xml': 30,
           'criminal/feed-rss.xml': 16,
-          'themoth/feed-rss.xml': 16,
+          'themoth/feed-rss.xml': 30,
```
cavis, line 954: 284 is a test show, so no need to scale on that.  (And especially no need to scale in past the minimum).

TTD/Moth should be scaled out to the max when the private feeds change.

farski (approved):

PR #648, opened 2022-10-11
gsuite subdomain for The World email migration

klbaradan:
gsuite.theworld.org pointed to PRX Gmail

(done correct this time)


farski (approved):
Looks right. Feel free to merge, but note that merging will not automatically deploy the change for this particular domain, so let me know when it needs to be live and I can run the update manually.

PR #650, opened 2022-10-12
Update theworld.org-hosted-zone.yml

klbaradan:
MS TXT verification for o365.theworld.org


klbaradan:
please publish once I accept it...

PR #651, opened 2022-10-13
Image uploads

cavis:
For PRX/id.prx.org#276.

- Adds a new S3 bucket, for ID temporary image uploads
- Configure for Porter image processing/callbacks
- Remove CMS references (already gone from ID codebase)

### DIFF
--- a/stacks/apps/id.yml
+++ b/stacks/apps/id.yml
@@ -31,10 +31,10 @@ Parameters:
   AmazonSesSmtpCredentialsGeneratorServiceToken: { Type: String }
   CloudWatchAlarmTaggerServiceToken: { Type: String }
   CloudWatchLogGroupTaggerServiceToken: { Type: String }
-  CmsHostname: { Type: String }
   ExchangeHostname: { Type: String }
   IdHostname: { Type: String }
   MediajointS3BucketArn: { Type: AWS::SSM::Parameter::Value<String> }
+  PorterJobExecutionSnsTopicArn: { Type: String }
   SharedAuroraMysqlEndpoint: { Type: String }
   SharedAuroraMysqlPort: { Type: String }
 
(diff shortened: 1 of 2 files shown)


stacks/apps/id.yml
```diff
@@ -366,6 +396,48 @@ Resources:
...
+        RestrictPublicBuckets: true
+      Tags:
+        - { Key: prx:meta:tagging-version, Value: "2021-04-07" }
+        - { Key: prx:cloudformation:stack-name, Value: !Ref AWS::StackName }
+        - { Key: prx:cloudformation:stack-id, Value: !Ref AWS::StackId }
+        - { Key: prx:cloudformation:root-stack-name, Value: !Ref RootStackName }
+        - { Key: prx:cloudformation:root-stack-id, Value: !Ref RootStackId }
+        - { Key: prx:ops:environment, Value: !Ref EnvironmentType }
+        - { Key: prx:dev:application, Value: ID }
+
```
cavis, line 440: Is there any other bucket config necessary, to give Porter read-access?  Trying to remember if that was a manual thing, after the bucket is created...

stacks/apps/id.yml
```diff
@@ -366,6 +396,48 @@ Resources:
       SecretAccessKey: !GetAtt SesSendingAccessKey.SecretAccessKey
       Region: !Ref AWS::Region
 
+  IdUploadsBucket:
+    Type: AWS::S3::Bucket
+    # TODO: necessary?
+    # DeletionPolicy: Retain
+    # UpdateReplacePolicy: Retain
```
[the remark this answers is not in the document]
farski: CloudFormation will try to delete a bucket if it's not retained if the resource disappears from the template. S3 will throw an error if you try to delete a bucket that still has any objects in it. AFAIK there's no way to get CFN to empty a bucket before deleting it (without custom resources, and even then that's pretty scary). So retention is added to simply avoid the issue, and leave bucket cleanup to the user, since it's hard to guarantee that a bucket will be empty when CFN wants to delete it.

stacks/apps/id.yml
farski: For a bucket in the same account as Porter, no bucket policies are required (due to the fact that for resources in the same account, you only need to grant access on one side. Porter has read access to `*`, so that satisfies the permissions for all buckets in the same account. For a bucket in a different account, permission would have to be granted explicitly on both sides, so a bucket policy would be required)

farski (changes requested):

farski (approved):

PR #654, opened 2022-10-18
Add new metrics filter and alarm on the slow log group

svevang:


stacks/apps/augury.yml
```diff
@@ -739,6 +750,20 @@ Resources:
         - { Key: prx:ops:cloudwatch-log-group-name, Value: !Ref WorkerTaskLogGroup }
         - { Key: prx:dev:family, Value: Dovetail }
         - { Key: prx:dev:application, Value: Augury }
+  AugurySlowJobsProducedErrorsAlarm:
```
svevang, line 753: Is there a functional approach here?  Something like a template include that defines all the alarms / filters and takes in an environment and log group ARN as parameters?

cavis: Hmm... I can see that - but I'm also fine with a bit of code duplication in this case.  CFN templates are pretty verbose to begin with!

stacks/apps/augury.yml
```diff
@@ -739,6 +750,20 @@ Resources:
...
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      AlarmName: !Sub ERROR [Augury] Worker <${EnvironmentTypeAbbreviation}> IS THROWING THE ERRORS (${RootStackName})
+      AlarmDescription: !Sub >-
+        ${EnvironmentType} Augury's worker task has raised an error. Check the logs!
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      EvaluationPeriods: 3
+      MetricName: !GetAtt Constants.JobErrorsCountMetricName
```
cavis, line 761: Should this be `SlowJobErrorsCountMetricName`?

PR #660, opened 2022-11-09
Dovetail traffic lambda secret

cavis:
Upcoming changes will cut off the traffic lambda from generating fake downloads on our staging/demo podcasts.

This wires up a shared secret to continue allowing it for our demo/test podcasts.

### DIFF
--- a/stacks/apps/dovetail-router.yml
+++ b/stacks/apps/dovetail-router.yml
@@ -843,6 +843,8 @@ Resources:
               ValueFrom: !Sub /prx/${EnvironmentTypeAbbreviation}/Spire/Dovetail-Router/augury-token
             - Name: FEEDER_TOKEN
               ValueFrom: !Sub /prx/${EnvironmentTypeAbbreviation}/Spire/Dovetail-Router/feeder-token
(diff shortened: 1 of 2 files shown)


stacks/apps/dovetail-router.yml
```diff
@@ -843,6 +843,8 @@ Resources:
               ValueFrom: !Sub /prx/${EnvironmentTypeAbbreviation}/Spire/Dovetail-Router/augury-token
             - Name: FEEDER_TOKEN
               ValueFrom: !Sub /prx/${EnvironmentTypeAbbreviation}/Spire/Dovetail-Router/feeder-token
+            - Name: TRAFFIC_SECRET
+              ValueFrom: /prx/global/Spire/Dovetail-Router/traffic-secret
```
cavis, line 847: Good enough name?  As there's only 1 traffic lambda, this needs to be shared in every environment and region.

kookster: I think that's good!

kookster (approved):

farski (approved):

PR #661, opened 2022-11-17
Add function template

svevang:
Adds the api bridge lambda using the S3 uploader as an example


stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+  CloudWatchAlarmTaggerServiceToken: { Type: String }
+  CloudWatchLogGroupTaggerServiceToken: { Type: String }
+  CodeS3Bucket: { Type: String }
+  CodeS3ObjectKey: { Type: AWS::SSM::Parameter::Value<String> }
+
+Resources:
+  HttpApi:
+    Type: AWS::Serverless::HttpApi
+    Properties:
+      StageName: !Ref EnvironmentType
```
farski, line 25: Including a stage name here is not necessary when using Serverless resources, and somewhat duplicative since the entire deployment is for a given environment type. The resulting URLs are simpler if it's excluded, so I would drop it.

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+        prx:meta:tagging-version: "2021-04-07"
+        prx:cloudformation:stack-name: !Ref AWS::StackName
+        prx:cloudformation:stack-id: !Ref AWS::StackId
+        prx:cloudformation:root-stack-name: !Ref RootStackName
+        prx:cloudformation:root-stack-id: !Ref RootStackId
+        prx:ops:environment: !Ref EnvironmentType
+        prx:dev:family: Dovetail
+        prx:dev:application: Apple Api Bridge
+
+  AppleApiBridgeLambdaFunction:
```
farski, line 38: Our newer CloudFormation templates have moved to a slightly less verbose naming convention for some resources, so this would be `AppleApiBridgeFunction`. Purely cosmetic in the code, but would be more inline with most other templates.

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+      Architectures:
+        - x86_64
+      Events:
+        ApiRequest:
+          Properties:
+            ApiId: !Ref HttpApi
+            Method: post
+            Path: /post
+          Type: HttpApi
+        Bridge:
```
farski, line 57: I think you'll want this to be an `HttpApi` event type, like `ApiRequest`, and include the `ApiId`. That assumes both of these paths are being handled by the HTTP API Gateway.

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+      Tags:
+        prx:meta:tagging-version: "2021-04-07"
+        prx:cloudformation:stack-name: !Ref AWS::StackName
+        prx:cloudformation:stack-id: !Ref AWS::StackId
+        prx:cloudformation:root-stack-name: !Ref RootStackName
+        prx:cloudformation:root-stack-id: !Ref RootStackId
+        prx:ops:environment: !Ref EnvironmentType
+        prx:dev:family: Dovetail
+        prx:dev:application: Apple Api Bridge
+
```
farski, line 71: I'd remove this newline

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+  AppleApiBridgeLambdaLogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: !Sub /aws/lambda/${AppleApiBridgeLambdaFunction}
+      RetentionInDays: 60
+
```
farski, line 77: I'd remove this newline

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+        - { Key: prx:meta:tagging-version, Value: "2021-04-07" }
+        - { Key: prx:cloudformation:stack-name, Value: !Ref AWS::StackName }
+        - { Key: prx:cloudformation:stack-id, Value: !Ref AWS::StackId }
+        - { Key: prx:cloudformation:root-stack-name, Value: !Ref RootStackName }
+        - { Key: prx:cloudformation:root-stack-id, Value: !Ref RootStackId }
+        - { Key: prx:ops:environment, Value: !Ref EnvironmentType }
+        - { Key: prx:dev:family, Value: Dovetail }
+        - { Key: prx:dev:application, Value: Apple Api Bridge }
+
+  BridgeLambdaElevatedThrottleAlarm:
```
farski, line 163: If the function is renamed, this would become `AppleApiBridgeFunctionElevatedThrottleAlarm`

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+        - Name: FunctionName
+          Value: !Ref AppleApiBridgeLambdaFunction
+      EvaluationPeriods: 1
+      MetricName: Throttles
+      Namespace: AWS/Lambda
+      Period: 60
+      Statistic: Sum
+      Threshold: 0
+      TreatMissingData: notBreaching
+  BridgeLambdaElevatedThrottleAlarmTags:
```
farski, line 181: If the function is renamed, this would become `AppleApiBridgeFunctionElevatedThrottleAlarmTags`

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+        - { Key: prx:meta:tagging-version, Value: "2021-04-07" }
+        - { Key: prx:cloudformation:stack-name, Value: !Ref AWS::StackName }
+        - { Key: prx:cloudformation:stack-id, Value: !Ref AWS::StackId }
+        - { Key: prx:cloudformation:root-stack-name, Value: !Ref RootStackName }
+        - { Key: prx:cloudformation:root-stack-id, Value: !Ref RootStackId }
+        - { Key: prx:ops:environment, Value: !Ref EnvironmentType }
+        - { Key: prx:dev:family, Value: Dovetail }
+        - { Key: prx:dev:application, Value: Apple Api Bridge }
+
+  BridgeLambdaElevatedVolumeAlarm:
```
farski, line 197: If the function is renamed, this would become `AppleApiBridgeFunctionElevatedVolumeAlarm`

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+          Value: !Ref AppleApiBridgeLambdaFunction
+      DatapointsToAlarm: 2
+      EvaluationPeriods: 3
+      MetricName: Invocations
+      Namespace: AWS/Lambda
+      Period: 60
+      Statistic: Sum
+      Threshold: 500
+      TreatMissingData: notBreaching
+  BridgeLambdaElevatedVolumeAlarmTags:
```
farski, line 218: If the function is renamed, this would become `AppleApiBridgeFunctionElevatedVolumeAlarmTags`

stacks/apps/apple-api-bridge.yml
farski, line 197: Invocation volume isn't something we alarm on most Lambda functions (it exists in the S3 Signing template only because I use that template as a test bed for some CloudFormation things). No real harm in keeping it, but not necessary.

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+        - { Key: prx:meta:tagging-version, Value: "2021-04-07" }
+        - { Key: prx:cloudformation:stack-name, Value: !Ref AWS::StackName }
+        - { Key: prx:cloudformation:stack-id, Value: !Ref AWS::StackId }
+        - { Key: prx:cloudformation:root-stack-name, Value: !Ref RootStackName }
+        - { Key: prx:cloudformation:root-stack-id, Value: !Ref RootStackId }
+        - { Key: prx:ops:environment, Value: !Ref EnvironmentType }
+        - { Key: prx:dev:family, Value: Dovetail }
+        - { Key: prx:dev:application, Value: Apple Api Bridge }
+
+  BridgeLambdaDangerousErrorRateAlarm:
```
farski, line 234: I would use either the error alarms from above, or this error rate alarm. Having both probably ends up being noisy (this, again, exists in the S3 Signing template more as a reference for how to do metric math than actual need)

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+      Tags:
+        - { Key: prx:meta:tagging-version, Value: "2021-04-07" }
+        - { Key: prx:cloudformation:stack-name, Value: !Ref AWS::StackName }
+        - { Key: prx:cloudformation:stack-id, Value: !Ref AWS::StackId }
+        - { Key: prx:cloudformation:root-stack-name, Value: !Ref RootStackName }
+        - { Key: prx:cloudformation:root-stack-id, Value: !Ref RootStackId }
+        - { Key: prx:ops:environment, Value: !Ref EnvironmentType }
+        - { Key: prx:dev:application, Value: Apple API Bridge }
+
+  BridgeLambdaCriticalErrorRateAlarm:
```
farski, line 288: Same as above: would pick either the rate alarms, or the volume alarms.

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
+# stacks/apps/apple-api-bridge.yml
```
farski, line 1: For consistency, this file should probably be called `dovetail-apple-api-bridge.yml`. Some of the oldest templates in the Dovetail family don't follow that pattern, but only because they predate it.

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+# 100A
```
farski, line 2: It doesn't look like this template has any hard dependencies on any other stacks, so it could go in the 100A template. But, if it had any soft/implicit dependencies, it could make sense to go in a higher-order template. Feeder is in 200A, so if it only ever really makes sense for the API Bridge to exist after Feeder, I'd probably put this in 300A. If it doesn't depend on Feeder (or anything else) in practice, it can stay in 100A.

Also worth noting that I don't see an addition to the 100A template in this PR. It will have to get added to one of those templates to actually get deployed.

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+    Properties:
+      CodeUri:
+        Bucket: !Ref CodeS3Bucket
+        Key: !Ref CodeS3ObjectKey
+      Handler: app.lambdaHandler
+      Runtime: nodejs18.x
+      Description: !Sub ${EnvironmentType} Apple API Bridge
+      MemorySize: 128
+      Architectures:
+        - x86_64
```
farski, line 49: If there are no binary dependencies for this app, I would use the `arm64` architecture. It tends to be more resource efficient.

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+        - { Key: prx:dev:application, Value: Apple Api Bridge }
+
+Outputs:
+  ApiUrl:
+    Description: URL of the Apple API Bridge API
+    Value:
+      Fn::Sub: "https://${HttpApi}.execute-api.${AWS::Region}.${AWS::URLSuffix}/${EnvironmentType}/"
+  HttpApiId:
+    Description: Api id of HttpApi for Apple Api Bridge
+    Value:
```
farski, line 348: Would use the shorthand syntax: `Value: !Ref HttpApi`

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+        - { Key: prx:cloudformation:root-stack-name, Value: !Ref RootStackName }
+        - { Key: prx:cloudformation:root-stack-id, Value: !Ref RootStackId }
+        - { Key: prx:ops:environment, Value: !Ref EnvironmentType }
+        - { Key: prx:dev:family, Value: Dovetail }
+        - { Key: prx:dev:application, Value: Apple Api Bridge }
+
+Outputs:
+  ApiUrl:
+    Description: URL of the Apple API Bridge API
+    Value:
```
farski, line 344: Would use the shorthand syntax: `Value: !Sub https://${HttpApi}.execute-api.${AWS::Region}.${AWS::URLSuffix}/${EnvironmentType}/` (though `EnvironmentType` may be unnecessary if the `StageName` is removed above)

stacks/apps/apple-api-bridge.yml
```diff
@@ -0,0 +1,349 @@
...
+      Fn::Sub: "https://${HttpApi}.execute-api.${AWS::Region}.${AWS::URLSuffix}/${EnvironmentType}/"
+  HttpApiId:
```
farski, line 346: This value should get passed out of the middle template (100A, 200A, etc), and then, in the root stack, should get passed into the Dashboards stack, so that it can be added to the *Requests* graph. There are a bunch of examples in there already of other API Gateway-based metrics being added to that.

farski (changes requested):

stacks/apps-100A.yml
```diff
@@ -139,6 +139,30 @@ Resources:
...
+    Properties:
+      Parameters:
+        EnvironmentType: !Ref EnvironmentType
+        EnvironmentTypeAbbreviation: !Ref EnvironmentTypeAbbreviation
+        RootStackName: !Ref RootStackName
+        RootStackId: !Ref RootStackId
+        CloudWatchAlarmTaggerServiceToken: !Ref CloudWatchAlarmTaggerServiceToken
+        CloudWatchLogGroupTaggerServiceToken: !Ref CloudWatchLogGroupTaggerServiceToken
+        CodeS3Bucket: !Ref DeploymentPackageBucketName
+        CodeS3ObjectKey: !Sub /prx/${EnvironmentTypeAbbreviation}/Spire/Dovetail-Counts/pkg/s3-object-key
```
farski, line 153: Don't want `Dovetail-Counts` here

stacks/apps/dovetail-apple-api-bridge.yml
```diff
@@ -0,0 +1,198 @@
+# stacks/apps/apple-api-bridge.yml
```
farski, line 1: `stacks/apps/dovetail-apple-api-bridge.yml`

stacks/apps/dovetail-apple-api-bridge.yml
```diff
@@ -0,0 +1,198 @@
...
+    Properties:
+      CodeUri:
+        Bucket: !Ref CodeS3Bucket
+        Key: !Ref CodeS3ObjectKey
+      Handler: app.lambdaHandler
+      Runtime: nodejs18.x
+      Description: !Sub ${EnvironmentType} Apple API Bridge
+      MemorySize: 128
+      Architectures:
+        - x86_64
```
farski, line 48: Just confirming that we need x86

stacks/apps/dovetail-apple-api-bridge.yml
```diff
@@ -0,0 +1,198 @@
...
+        - { Key: prx:dev:application, Value: Apple Api Bridge }
+
+  # Lambda Alarms
+  AppleApiBridgeFunctionElevatedErrorAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: IsProduction
+    Properties:
+      AlarmName: !Sub WARN [Apple Api Bridge] Lambda function <${EnvironmentTypeAbbreviation}> INVOCATIONS ERRORS (${RootStackName})
+      AlarmDescription: !Sub >-
+        ${EnvironmentType} Apple API Bridge Lambda function is failing
```
farski, line 94: Is it possible to expand this description a bit to include some details on what the effects of this alarm. examples [here](https://github.com/PRX/Infrastructure/blob/main/stacks/apps/dovetail-counts.yml#L160) and [here](https://github.com/PRX/Infrastructure/blob/main/stacks/apps/dovetail-analytics.yml#L113), docs [here](https://github.com/PRX/internal/wiki/AWS:-CloudWatch-Alarms#description)

stacks/apps/dovetail-apple-api-bridge.yml
```diff
@@ -0,0 +1,198 @@
...
+        - { Key: prx:dev:family, Value: Dovetail }
+        - { Key: prx:dev:application, Value: Apple Api Bridge }
+
+  AppleApiBridgeFunctionDangerousErrorAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: IsProduction
+    Properties:
+      AlarmName: !Sub ERROR [Apple Api Bridge] Lambda function <${EnvironmentTypeAbbreviation}> INVOCATIONS ERRORS (${RootStackName})
+      AlarmDescription: !Sub >-
+        ${EnvironmentType} Apple Api Bridge Lambda is failing a lot
```
farski, line 128: Expound if possible

stacks/apps/dovetail-apple-api-bridge.yml
```diff
@@ -0,0 +1,198 @@
...
+        - { Key: prx:dev:family, Value: Dovetail }
+        - { Key: prx:dev:application, Value: Apple Api Bridge }
+
+  AppleApiBridgeFunctionElevatedThrottleAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: IsProduction
+    Properties:
+      AlarmName: !Sub ERROR [Apple Api Bridge] Lambda function <${EnvironmentTypeAbbreviation}> INVOCATIONS BEING THROTTLED (${RootStackName})
+      AlarmDescription: !Sub >-
+        ${EnvironmentType} Apple Api Bridge Lambda function is being throttled.
```
farski, line 163: Expound if possible

stacks/dashboards.yml
```diff
@@ -158,6 +159,7 @@ Resources:
                           [ "AWS/ApplicationELB", "RequestCount", "TargetGroup", "${CastleTargetGroupFullName}", "LoadBalancer", "${SharedAlbFullName}", { "label": "Castle" } ],
                           [ "AWS/ApplicationELB", "RequestCount", "TargetGroup", "${CmsTargetGroupFullName}", "LoadBalancer", "${SharedAlbFullName}", { "label": "CMS" } ],
                           [ "AWS/ApplicationELB", "RequestCount", "TargetGroup", "${DovetailTargetGroupFullName}", "LoadBalancer", "${DovetailAlbFullName}", { "label": "Dovetail-Router" } ],
+                          [ "AWS/ApiGateway", "Count", "ApiId", "${DovetailAppleApiBridgeApiId}", { "label": "Feeder Apple Api Bridge" } ],
```
farski, line 162: Maybe `Dovetail-Apple API Bridge` for the label, to match other places

stacks/dashboards.yml
```diff
@@ -43,6 +43,7 @@ Parameters:
 
   DovetailAlbName: { Type: String }
   DovetailAlbFullName: { Type: String }
+  DovetailAppleApiBridgeApiId: { Type: String }
```
farski, line 46: This line should stand on its own, surrounded by whitespace. Each distinct app gets its own little section.

stacks/root.yml
```diff
@@ -875,6 +876,7 @@ Resources:
 
         DovetailAlbName: !GetAtt Apps300AStack.Outputs.DovetailAlbName
         DovetailAlbFullName: !GetAtt Apps300AStack.Outputs.DovetailAlbFullName
+        DovetailAppleApiBridgeApiId: !GetAtt Apps100AStack.Outputs.DovetailAppleApiBridgeApiId
```
farski, line 879: This line should stand on its own, surrounded by whitespace. Each distinct app gets its own little section.

farski (changes requested):

farski (approved):
I think this looks good. One last change would be, in the dashboard stack and its params in the root stack, to make sure that the `DovetailTargetGroupFullName` doesn't get orphaned from `DovetailAlbFullName` and `DovetailAlbName`. Those three should be grouped together, and then `DovetailAppleApiBridgeApiId` should be I guess before them alphabetically.

PR #662, opened 2022-11-28
Split kinesis batch on error

cavis:
Closes PRX/internal#873.

Manually trying this in us-west-2 first.


cavis:
West-2 looks fine for the past few days.  (No noticeable difference in cloudwatch).

Onward!

PR #664, opened 2022-12-08
Truncate shipped IPv6s

cavis:
Per some IAB2.1...

We should only use half of IPv6 when calculating listener_ids.  But continue using all of IPv4.

While I was at it - also truncated the shipped IPv6 IPs / XFFs the same way.

Should probably test this out carefully before actually deploying.

### DIFF
--- a/etc/cdn-log-shipper/log-shipper.yml
+++ b/etc/cdn-log-shipper/log-shipper.yml
@@ -92,7 +92,6 @@ Resources:
         const crypto = require('crypto');
 
         const IPV4_MASK = /\.[0-9]{1,3}$/;
-        const IPV6_MASK = /:[0-9a-fA-F]{0,4}$/;
         const maskIp = (ip, field) => {
           if (ip.match(IPV4_MASK)) {
             return ip.replace(IPV4_MASK, '.0');
--- a/etc/cdn-log-shipper/log-shipper.yml
+++ b/etc/cdn-log-shipper/log-shipper.yml
@@ -100,7 +100,7 @@ Resources:
             // take up to 4 chunks - drop the rest
             const chunks = ip.split(':').slice(0, 4);
 
-            # remove all trailing ':', then add '::' (zeroes)
+            // remove all trailing ':', then add '::' (zeroes)
             return chunks.join(':').replace(/:+$/, '') + '::'
           } else if (ip === '-') {
             // that's ok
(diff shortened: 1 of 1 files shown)


etc/cdn-log-shipper/log-shipper.yml
```diff
@@ -92,14 +92,15 @@ Resources:
...
         const maskIp = (ip, field) => {
           if (ip.match(IPV4_MASK)) {
             return ip.replace(IPV4_MASK, '.0');
-          } else if (ip.endsWith(':')) {
-            return ip;
-          } else if (ip.match(IPV6_MASK)) {
-            return ip.replace(IPV6_MASK, ':0');
+          } else if (ip.includes(':')) {
+            // take up to 4 chunks - drop the rest
+            const chunks = ip.split(':').slice(0, 4);
```
cavis, line 100: It seemed like that old regex didn't _really_ work for all cases.

This one is better, but I suspect it doesn't handle some edge cases where there are `::` in the middle of the string somewhere.

etc/cdn-log-shipper/log-shipper.yml
```diff
@@ -151,10 +152,17 @@ Resources:
 
             // calculate listener_ids
             datas.forEach(data => {
-              const literalIp = data['x-forwarded-for'] || data['c-ip'] || '';
+              // use leftmost XFF or IP
+              const xffParts = (data['x-forwarded-for'] || '').split(',').map(s => s.trim()).filter(s => s);
+              const leftMostIp = xffParts[0] || data['c-ip'];
```
cavis, line 157: Note that DTR does some other logic here.  It actually bothers to parse/verify the XFF pieces, and removes loopbacks/private-nets, etc.

In reality, I doubt it will make much of a difference?  This less smart code should do fine.

kookster: I think that's fine

etc/cdn-log-shipper/log-shipper.yml
cavis: So your example expands to `1234:5678:0:0:0:0:abcd:efgh`.  Ideally we'd truncate that to `1234:5678::` or equivalent.

Looks like my code would end up with 2 `::` in there - `1234:5678::abcd::`.  Which isn't valid (it's ambiguous how many 0s go where).

In reality, I'm doubtful we'd ever see that?  As zeroes tend to be towards the end of the IPv6.  Trying to think of how to handle without an actual IP parsing library...

kookster: I think this is good enough, and if we need to bring in a lib, well, then we will!

kookster (approved):
lgtm!

PR #666, opened 2022-12-29
Drop http api in favor of lambda url

svevang:
Looking at the upload functionality in feeder, I noticed that the http api gateway trigger has a fixed timeout and does not reflect the underlying lambda's timeout.

Digging a bit deeper, it looks like the lambda url resource on the lambda itself can be used to invoke the lambda directly and will honor the underlying timeout.

This PR:

- sets up a new lambda url resource on the lambda
- removes the api gateway trigger
- patches that new url through the stacks in place of the api gateway url

### DIFF
--- a/stacks/apps-100A.yml
+++ b/stacks/apps-100A.yml
@@ -476,9 +476,9 @@ Outputs:
     Value: !GetAtt CmsStack.Outputs.TargetGroupFullName
 
   DovetailAppleApiBridgeEndpointUrl:
-    Value: !GetAtt DovetailAppleApiBridgeStack.Outputs.ApiUrl
-  DovetailAppleApiBridgeApiId:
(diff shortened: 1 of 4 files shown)


farski:
@svevang I don't think this removes the API Gateway, or the event trigger on the Lambda for the API Gateway.

svevang:
@farski got it thanks for the review.  Can you take another look?

svevang:
Functionally, it's very very similar. I gather that the lambda URLs invoke the url similarly to how the AWS SDKs would.  so unlike the Http api, there's no differentiation between GET/POST etc and all path components in the URL are treated the same (`/foo` is the same as `/`). but yeah, functionally this should be the same, and allow the lambda to do some longer upload work.

Issue #659, opened 2022-11-04
Augury alarm on error log levels


svevang:
Set up a new alarm on augury log messages that are `:error` log level.  In Augury's Ougai logger these are numeric values.

PR #667, opened 2023-01-06
Set up alarms on error log level messages

svevang:
Solves #659

### DIFF
--- a/stacks/apps/augury.yml
+++ b/stacks/apps/augury.yml
@@ -81,6 +81,10 @@ Resources:
       SlowJobErrorsCountMetricName: !Sub SlowJobErrorsCountNewCluster${EnvironmentType}
       TargetingRecordCacheSizeMetricsName: !Sub TargetingRecordCacheSize${EnvironmentType}
       AdFilesS3BucketName: !Select [5, !Split [":", !Ref AdFilesS3BucketArn]]
+      WorkerLoggedErrorsMetricName: !Sub WorkerLoggedErrors${EnvironmentType}
+      SlowWorkerLoggedErrorsMetricName: !Sub SlowWorkerLoggedErrors${EnvironmentType}
+      WebLoggedErrorsMetricName: !Sub WebLoggedErrors${EnvironmentType}
+
 
   RedisSecurityGroup:
     Type: AWS::EC2::SecurityGroup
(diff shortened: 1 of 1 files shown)


svevang:
Before this lands we need to handle (on augury):

- ActionController::RoutingError
- HalApi::Errors::NotFound

Regular 404's are being handled (it seems) so make ignore these two error classes in the augury code. So downgrade the log level for these errors from FATAL (level: 60) to INFO (level: 30).

stacks/apps/augury.yml
```diff
@@ -677,6 +680,78 @@ Resources:
         - MetricName: !GetAtt Constants.SlowJobErrorsCountMetricName
           MetricNamespace: !Ref kMetricFilterNamespace
           MetricValue: "1"
+  WebLoggedErrorsMetricFilter:
+    Type: AWS::Logs::MetricFilter
+    Properties:
+      FilterPattern: |-
+        { $.level >= 50 }
```
kookster, line 687: So this is where the numeric number value gets checked - nice!

stacks/apps/augury.yml
```diff
@@ -677,6 +680,78 @@ Resources:
...
+        - MetricName: !Sub Constants.SlowWorkerLoggedErrorsMetricName
+          MetricNamespace: !Ref kMetricFilterNamespace
+          MetricValue: "1"
+  AuguryWebLoggedErrorsAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      AlarmName: !Sub ERROR [Augury] Web <${EnvironmentTypeAbbreviation}> IS LOGGING ERRORS (${RootStackName})
+      AlarmDescription: !Sub >-
+        ${EnvironmentType} Augury's web task has logged an error. Check the logs!
+      ComparisonOperator: GreaterThanOrEqualToThreshold
```
kookster, line 719: I see in other error alarms we use `ComparisonOperator: GreaterThanThreshold`, this seems right here though, as it is consistent with other Augury alarms, and it will keep this from being a flood of slack messages, while the low threshold will alert on any error.

kookster (approved):
lgtm!

stacks/apps/augury.yml
```diff
@@ -81,6 +81,9 @@ Resources:
       SlowJobErrorsCountMetricName: !Sub SlowJobErrorsCountNewCluster${EnvironmentType}
       TargetingRecordCacheSizeMetricsName: !Sub TargetingRecordCacheSize${EnvironmentType}
       AdFilesS3BucketName: !Select [5, !Split [":", !Ref AdFilesS3BucketArn]]
+      WorkerLoggedErrorsMetricName: !Sub WorkerLoggedErrors${EnvironmentType}
+      SlowWorkerLoggedErrorsMetricName: !Sub SlowWorkerLoggedErrors${EnvironmentType}
+      WebLoggedErrorsMetricName: !Sub WebLoggedErrors${EnvironmentType}
```
farski, line 86: I believe CloudFormation now supports dimensions on metric filters. This should allow us to have a single metric name (e.g., `WebLoggedErrors`), and something like an `Environment` dimension whose value is `${EnvironmentType}`, getting us away from having to make these kind of gross metric names that cram the env type in there, and no longer need to create parameters for the names.

I haven't had a chance to try it, so there's a chance the docs have been updated but the functionality isn't actually supported yet, but this seems like a good chance to test.

The alarms would also have to be updated to add a dimension.

svevang: I went down the road a bit and hit a stumbling block with the metric filter dimensions.  It seems that the dimensions are derived from the log line itself.  [The `Dimension` property on the metric filter](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-logs-metricfilter-dimension.html) has this caveat:

```
This dimension will only be published for a metric if the value is found in the log event.
```

Which makes me think that a "constant" value would somehow try to match the whole log line, or something else unexpected.  Still worth pursuing?

farski: That's annoying. There may be some clever way to inject those values using the filter expression, but I'd have to mess around with it to see. Let's punt on this for now.

farski (approved):

PR #669, opened 2023-01-23
Fix/augury logged error alarms

svevang:
Fixes a set of alarms that don't work for a couple reasons:

- Too many evaluation periods to likely ever satisfy
- Incorrectly access the `Constants` resource attributes

This PR fixes the above, and switches the style to match the 500's alarm:

- Use a single evaluation period
- Use `GreaterThanThreshold`

In the modified alarms here, the alarm is considered existential: if the error occurs, then set off the alarm.  Re: Multiple evaluation periods, I think only in a self-healing scenario (like the actuals falling behind or forecasts slow down a bit) would we want to ignore the error metrics for a single evaluation period.  These alarms imply something has gone wrong that needs outside intervention.

### DIFF
--- a/stacks/apps/augury.yml
+++ b/stacks/apps/augury.yml
@@ -716,13 +716,13 @@ Resources:
       AlarmName: !Sub ERROR [Augury] Web <${EnvironmentTypeAbbreviation}> IS LOGGING ERRORS (${RootStackName})
       AlarmDescription: !Sub >-
         ${EnvironmentType} Augury's web task has logged an error. Check the logs!
-      ComparisonOperator: GreaterThanOrEqualToThreshold
-      EvaluationPeriods: 3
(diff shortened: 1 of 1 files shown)


farski (approved):

stacks/apps/augury.yml
```diff
@@ -687,7 +687,7 @@ Resources:
         { ($.level >= 50) && ($.level < 60) }
       LogGroupName: !Ref WebTaskLogGroup
       MetricTransformations:
-        - MetricName: !Sub Constants.WebLoggedErrorsMetricName
+        - MetricName: !GettAtt Constants.WebLoggedErrorsMetricName
```
farski, line 690: Two Ts

farski (changes requested):

PR #670, opened 2023-01-25
Feat/tag augury logged error alarms

svevang:
Fixes up the previous #669 

This:

- adds the tags to the "logged errors" alarms
- reformats so the metric filters, alarms and tags are cuddled together

### DIFF
--- a/stacks/apps/augury.yml
+++ b/stacks/apps/augury.yml
@@ -704,6 +704,22 @@ Resources:
       Statistic: Sum
       Threshold: 0
       TreatMissingData: notBreaching
+  AuguryWebLoggedAlarmTags:
+    Type: Custom::CloudWatchAlarmTags
+    Condition: EnableWorkers # See README
+    Properties:
+      ServiceToken: !Ref CloudWatchAlarmTaggerServiceToken
+      AlarmArn: !GetAtt AuguryWebLoggedErrorsAlarm.Arn
+      Tags:
+        - { Key: prx:meta:tagging-version, Value: "2021-04-07" }
+        - { Key: prx:cloudformation:stack-name, Value: !Ref AWS::StackName }
+        - { Key: prx:cloudformation:stack-id, Value: !Ref AWS::StackId }
+        - { Key: prx:cloudformation:root-stack-name, Value: !Ref RootStackName }
+        - { Key: prx:cloudformation:root-stack-id, Value: !Ref RootStackId }
+        - { Key: prx:ops:environment, Value: !Ref EnvironmentType }
+        - { Key: prx:ops:cloudwatch-log-group-name, Value: !Ref WebTaskLogGroup }
+        - { Key: prx:dev:family, Value: Dovetail }
+        - { Key: prx:dev:application, Value: Augury }
   WorkersLoggedErrorsMetricFilter:
     Type: AWS::Logs::MetricFilter
     Properties:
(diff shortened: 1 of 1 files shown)


stacks/apps/augury.yml
```diff
@@ -738,6 +744,32 @@ Resources:
       Statistic: Sum
       Threshold: 0
       TreatMissingData: notBreaching
+  AuguryWorkerLoggedAlarmTags:
```
farski, line 747: `AuguryWorkerLoggedErrorsAlarmTags`

stacks/apps/augury.yml
```diff
@@ -738,6 +744,32 @@ Resources:
...
+    Type: Custom::CloudWatchAlarmTags
+    Condition: EnableWorkers # See README
```
farski, line 749: No condition on alarm

stacks/apps/augury.yml
```diff
@@ -752,6 +784,22 @@ Resources:
       Statistic: Sum
       Threshold: 0
       TreatMissingData: notBreaching
+  AugurySlowWorkerLoggedAlarmTags:
```
farski, line 787: `AugurySlowWorkerLoggedErrorsAlarmTags`

stacks/apps/augury.yml
```diff
@@ -752,6 +784,22 @@ Resources:
...
+    Type: Custom::CloudWatchAlarmTags
+    Condition: EnableWorkers # See README
```
farski, line 789: No condition on alarm

farski (changes requested):

stacks/apps/augury.yml
```diff
@@ -724,6 +704,32 @@ Resources:
       Statistic: Sum
       Threshold: 0
       TreatMissingData: notBreaching
+  AuguryWebLoggedAlarmTags:
```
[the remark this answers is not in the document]
svevang: Thank you for the review!   I looked at the other names and opted to drop "Augury" from these resources names.  It already is in the augury stack etc, so no need for the appname prefix.  And less typing to boot!

svevang:
In the end I'd like the slack alarms to link to the log files.  I think this PR gets us a bit closer.

It seems to get the reference to the log lines, I'd have to add one of these: https://github.com/PRX/Infrastructure/blob/ed1556a0b7cc6a64eb8ec4a372f2cf43f981306f/devops/tooling/chat-ops/cloudwatch-alarms/src/slack/log-groups.js#L11
 to the alarm resources.    Maybe?

farski:
Yes, as long as the namespace that an alarm belongs to is in that list and the alarm includes the `prx:ops:cloudwatch-log-group-name` tag with an appropriate value, the Slack messages will include a link to CloudWatch Logs Insights for the period the alarm covers. If you need to add any namespaces to that list, I wouldn't worry about a PR; just add them.

stacks/apps/augury.yml
```diff
@@ -752,6 +782,21 @@ Resources:
       Statistic: Sum
       Threshold: 0
       TreatMissingData: notBreaching
+  SlowWorkerLoggedAlarmTags:
```
farski, line 785: `SlowWorkerLoggedErrorsAlarmTags`

farski (changes requested):

farski (approved):

PR #671, opened 2023-01-27
Cleanup Feeder ENVs

cavis:
I also sorted the ENVs, so probably easiest to review commit by commit.

Summary of changes:

- Removed `APP_NAME` ... that was an aws-secrets thing
- Removed `APP_ENV` ... also an aws-secrets thing
- Removed `DB_PORT_` vars - they were deprecated awhile back
- Removed `FIXER_CALLBACK_QUEUE_NAME` - Feeder doesn't use it.  It actually [interpolates it](https://github.com/PRX/feeder.prx.org/blob/main/app/models/task.rb#L58).
- Removed `LANG` - rails defaults to `en_US.UTF-8` anyways
- Removed `RAILS_SERVE_STATIC_FILES` - Feeder always serves static files now
- Removed `WORKER_PAUSE` - it already [defaults to 30](https://github.com/PRX/feeder.prx.org/blob/master/config/shoryuken.yml#L7)
- Removed `OTEL_` honeycomb stuff - unused.
- Removed `WORKER_DATABASE_POOL_SIZE`, and just set `DATABASE_POOL_SIZE` in the worker task def.  (I'm not sure why it had its own ENV?  Maybe for local development?)

### DIFF
--- a/stacks/apps/feeder.yml
+++ b/stacks/apps/feeder.yml
@@ -460,22 +460,16 @@ Resources:
             - web
           Cpu: !If [IsProduction, 200, 128]
           Environment:
-            - Name: RAILS_ENV
-              Value: !Ref EnvironmentTypeLowercase
+            - Name: ANNOUNCE_RESOURCE_PREFIX
+              Value: !Ref AnnounceResourcePrefix
             - Name: APPLE_API_BRIDGE_URL
               Value: !Ref DovetailAppleApiBridgeEndpointUrl
+            - Name: AWS_ACCOUNT_ID
+              Value: !Ref AWS::AccountId
             - Name: AWS_DEFAULT_REGION
               Value: !Ref AWS::Region
             - Name: AWS_REGION
               Value: !Ref AWS::Region
-            - Name: AWS_ACCOUNT_ID
-              Value: !Ref AWS::AccountId
-            - Name: POSTGRES_HOST
-              Value: !Ref SharedAuroraPostgresqlEndpoint
-            - Name: POSTGRES_PORT
-              Value: !Ref SharedAuroraPostgresqlPort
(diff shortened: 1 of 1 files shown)


stacks/apps/feeder.yml
```diff
@@ -460,46 +459,30 @@ Resources:
             - web
           Cpu: !If [IsProduction, 200, 128]
           Environment:
-            - Name: APP_NAME
-              Value: feeder
-            - Name: RAILS_ENV
```
kookster, line 465: (moved lower, not deleted)

stacks/apps/feeder.yml
```diff
@@ -508,24 +491,14 @@ Resources:
...
+              Value: !Ref EnvironmentTypeLowercase
             - Name: PRX_HOST
               Value: !Ref BetaHostname
-            - Name: RAILS_SERVE_STATIC_FILES
-              Value: enabled
-            - Name: WORKER_PAUSE
-              Value: "30"
-            - Name: WORKER_DATABASE_POOL_SIZE
-              Value: !If [IsProduction, 16, 5]
-            - Name: WORKER_COUNT
```
kookster, line 519: Don't we need this still? I see it in the `config/shoryuken.yml`:
```
concurrency: <%= ENV['WORKER_COUNT'] || 25 %>
```

kookster (changes requested):
Don't we still need `WORKER_COUNT`?

stacks/apps/feeder.yml
cavis: Yep - still defined in the worker task def.  Just not the web.

kookster (approved):
nvm, approved!

PR #674, opened 2023-03-03
Add NEW_RELIC_NAME to Exchange envars

farski:
I don't know if this is why Exchange stopped reporting to NR, but it's the only app that didn't have this name configured as an envar. Maybe that's required now? I doubt it, but seems good to have here anyway.


cavis (approved):
👍

PR #679, opened 2023-04-10
Upgrade Feeder auth runtime

farski:


cavis (approved):
+1.

(Someday, we probably want to convert this to the new node streaming-responses.  Just in case the RSS is too large for the normal kind).

Issue #684, opened 2023-05-05
Log shipper for triton


kookster:
hearing from Jay Green that he's looking at using triton for audience analysis, and also to get into their ranker (we'd likely be in the top 10 if we do).

I haven't got the proposal yet, but sounds like a greater than 50% chance on this.

I believe this would be like magellan, and we'd add another log shipper configuration to go to a different s3 destination.

kookster:
got more docs on this!
https://userguides.tritondigital.com/mea/pcmmg/podcast-log-req.html

kookster:
this is still going through contracting - pushed it out a couple of sprints

kookster:
Hearing from Jay @ Soundrise that they expect to get the contract done, and then try to get this live with Triton in June or July

kookster:
Did a quick compare against a sample file
We don't have a hashed ip, just the listenerid
We have the range in 2 fields (start, end) not 1

So we'll go over those and see what might need to change

farski:
Logs are being sent to (our) S3 bucket `prx-triton-cdn-logs`, and Triton should have access. No further work needed on our end, unless they ask for changes.

PR #693, opened 2023-07-18
Add hashed ip for Triton

kookster:
For #684 
Not totally sure how to test this, but using a local script this seems to work ~, adding a single column for `byte-range`~
actually, they are fine with the separate start and end columns, they did ask for a hashed ip, so adding that instead.

*Don't merge until we get confirmation they want the truncated ipv6 used for the hashed ip*

### DIFF
--- a/etc/cdn-log-shipper/log-shipper.yml
+++ b/etc/cdn-log-shipper/log-shipper.yml
@@ -171,6 +171,15 @@ Resources:
             });
             fields.push('prx-listener-id');
 
+            datas.forEach(data => {
+              if (data['sc-range-start'].match(/^[0-9]+$/)) {
+                data['byte-range'] = `${data['sc-range-start']}-${data['sc-range-end']}`;
+              } else {
+                data['byte-range'] = '-';
+              }
+            });
+            fields.push('byte-range');
+
             // mask IP addresses
             datas.forEach(data => {
               data['c-ip'] = maskIp(data['c-ip'], 'c-ip');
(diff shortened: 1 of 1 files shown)


farski:
This is probably a better PR for @cavis, I've never looked at this code before. Seems reasonable to me though.

cavis (approved):
👍 seems good to me.

We may want to alert existing consumers of these logs to this change?  You'd hope they would be parsing the headers of the tsv.  But who knows.  Looks like you put this one last, so maybe fine.

etc/cdn-log-shipper/log-shipper.yml
```diff
@@ -171,6 +171,15 @@ Resources:
             });
             fields.push('prx-listener-id');
 
+            datas.forEach(data => {
+              if (data['sc-range-start'].match(/^[0-9]+$/)) {
+                data['byte-range'] = `${data['sc-range-start']}-${data['sc-range-end']}`;
+              } else {
+                data['byte-range'] = '-';
```
cavis, line 178: Would we ever hit this?

kookster: Yes, it only has numeric values for a 206 response, otherwise start and end are dashes

cavis: Oh right - this is the actual request Range header field.  Vs the "byte range the server sent to the client" we use for IAB2 calculations.

kookster:
heard back from Triton today - they don't need the byte range as a new field, they do need the ip hashed - will do that instead.

kookster:
triton also confirmed the ipv6 should be 64 bits (truncated) before hashing, like we do for the listenerid, so this is correct

PR #699, opened 2023-08-18
Move Spire templates

farski:
This is to complete a bit of reorganization that started a while ago, and pulls our main CloudFormation templates under the `spire/` directory.

This moves the files from `stacks/…` to `spire/templates/…`.

The key changes are:
1. Updating the `TemplateURL` used when creating change sets, which is hard-coded into the Lambda function that creates those change sets.
2. Updating which files from the Infrastructure repo zip file get copied to S3
3. Update the `TemplateURL` values for each nested stack in the `root.yml` template
4. For nested stacks that continue to pass the S3 URLs of these templates down to more deeply nested stacks, also update the values being passed into `TemplateUrlPrefix`

There's not a lot of risk in this change. If CloudFormation looks for a file and it doesn't exist (because I made a mistake in updating these paths), the deploy will fail. The only real risk is that somehow a stack will update to an actual file, but not the template intended for that stack, and it will start to replace all the resources. That seems basically impossible, since all the file names are hard coded, and I don't think any overlap, and they would also have to exactly share parameters.


spire/cd/README.md
```diff
@@ -98,9 +98,9 @@ The CloudFormation role has some serious implications that are worth being aware
...
 
-As an example, the [Publish](https://github.com/prx/publish.prx.org) app is managed through CD. It has its own [template](https://github.com/PRX/Infrastructure/blob/main/stacks/publish.prx.org.yml), which creates a number AWS resources necessary for deploying the app: log groups, ALB listener rules, ECS task definition and service, etc. The publish template is used with a nested stack resource in `root.yml`, so any time the root stack is updated, any changes that have been made to the publish template or its parameters will also be deployed.
+As an example, the [Exchange](https://github.com/prx/exchange.prx.org) app is managed through CD. It has its own [template](https://github.com/PRX/Infrastructure/blob/main/spire/templates/exchange.yml), which creates a number AWS resources necessary for deploying the app: log groups, ALB listener rules, ECS task definition and service, etc. The publish template is used with a nested stack resource in `root.yml`, so any time the root stack is updated, any changes that have been made to the publish template or its parameters will also be deployed.
```
cavis, line 103: Ah, Publish, already forgotten by the docs.

cavis (approved):
👍 looks good to me.

PR #701, opened 2023-09-06
Update mandatory SSO domains

farski:
I think this should be deployed to staging today late in the day, and we'll hold the prod deploy until 9 AM ET tomorrow


kookster (approved):
lgtm!

PR #704, opened 2023-09-18
Publicfeeds.net

cavis:
Adds the `PUBLIC_FEEDS_URL_PREFIX` env to the feeder stack.


farski (approved):
Looks good. If you wanted to be even more bulletproof, you could do the interpolation in the template's [constants](https://github.com/PRX/Infrastructure/blob/main/spire/templates/apps/feeder.yml#L72), to reduce the chances of the value getting updated in some places but not others.

cavis:
That `/f` path is such a Feeder-specific thing... I'm not sure why we'd use it elsewhere.  But if we do... agreed with interpolating it in one place!

PR #705, opened 2023-09-19
Setup shared app redis

cavis:
Moves the Augury stack redis to a shared stack, so both Augury and Feeder can use it.  (Both have similar evictions-allowed FIFO usage patterns).


spire/templates/shared-app-redis.yml
```diff
@@ -0,0 +1,107 @@
...
+        - { Key: prx:dev:application, Value: Common }
+      VpcSecurityGroupIds:
+        - !Ref SharedAppRedisSecurityGroup
+
+Outputs:
+  CacheName:
+    Value: !If [IsPrimaryRegion, !Ref SharedAppRedisCluster, !Ref "AWS::NoValue"]
+  CacheEndpointAddress:
+    Description: Cache endpoint hostname
+    Value: !If [IsPrimaryRegion, !GetAtt SharedAppRedisCluster.RedisEndpoint.Address, !Ref "AWS::NoValue"]
```
cavis, line 104: This cache only exists in the primary region - so hopefully this NoValue logic works to pass through templates?

The apps that use this ENV should handle it not being set, or being unavailable.

spire/templates/shared-app-redis.yml
```diff
@@ -0,0 +1,107 @@
...
+    Properties:
+      AutoMinorVersionUpgrade: true
+      CacheNodeType: !If
+        - IsProduction
+        - !Ref ProductionInstanceType
+        - !Ref StagingInstanceType
+      CacheSubnetGroupName: !Ref SharedAppRedisSubnetGroup
+      Engine: redis
+      EngineVersion: 6.x
+      NumCacheNodes: 1
```
cavis, line 86: I'm not sure we need > 1 nodes for these?  Augury has certainly been running without it.  We also don't need to back this up, or set any alarms (that I could think of).

Reading the [alarm best practices](https://aws.amazon.com/blogs/database/monitoring-best-practices-with-amazon-elasticache-for-redis-using-amazon-cloudwatch/) ... this work load is fine with evictions, so a memory % alarm doesn't make sense.  Additionally, these newer elasticache instances already reserve memory for system tasks (backups that we don't do, etc).  So I'm not sure we get anything monitoring swap usage.

farski (approved):

PR #715, opened 2023-10-12
Router scaling

### LINKED ISSUE #698: Dovetail router feed scaler not running in west-2
Noticed the feed scaler lambda has never run in west-2 (`FeedScalerFunction` in `dovetail-router.yml`). 

May be that event bridge events for `prx-feed` S3 just never make it to that other region.

Can we watch for changes to the replicated S3 bucket instead?  Or some other solution?

farski:
Closes #698

- Updates S3 bucket config (already deployed)
- Removes the bucket name from the EventBridge rule for triggering the scaler Lambda. This will now match on `Object Created` from any bucket, but we have EventBridge notifications turned on for very few buckets, and the likelihood of objects with these names being created in another bucket are pretty low. And even if we do get those false positives, we just scale up unnecessarily for a bit, which isn't really a problem.

### DIFF
--- a/spire/templates/apps/dovetail-router.yml
+++ b/spire/templates/apps/dovetail-router.yml
@@ -932,17 +932,17 @@ Resources:
               source:
                 - aws.s3
       Handler: index.handler
-      InlineCode: |
+      InlineCode: !Sub |
         const AWS = require('aws-sdk');
         const s3 = new AWS.S3();
-        const ddb = new AWS.DynamoDB({ region: 'us-east-1' });
-        const ecs = new AWS.ECS({ region: 'us-east-1' });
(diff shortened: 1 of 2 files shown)


spire/templates/apps/dovetail-router.yml
```diff
@@ -922,15 +922,10 @@ Resources:
           Properties:
             Pattern:
               detail:
-                bucket:
-                  name:
-                    - !Select [5, !Split [":", !Ref FeedsS3BucketArn]]
                 object:
                   key:
-                    - 815/feed-rss.xml
```
farski, line 930: This didn't seem to exist anymore

cavis (approved):
Should leave 815 in there (for staging), but otherwise 👍

cavis (approved):
👍

PR #718, opened 2023-10-23
Update hostnames for Augury and Feeder

farski:
- Change `augury.` to a redirect
- Unify staging and prod rules for Feeder
- Send `/api` and `/api/*` Feeder traffic directly to target group
- Redirect all other traffic to `feeder.` to `podcasts.dovetail`


cavis (approved):
👍

PR #719, opened 2023-10-24
Support change-password as ALB rule

farski:
Closes https://github.com/PRX/internal/issues/436

Gives us change-password support for all apps without having to build it in to each one.


cavis (approved):
👍

PR #725, opened 2023-11-13
Make sure the staging machines have 8gb

svevang:
Followup to #724 this bumps the ec2 image class to match (at least memory) what's in production.


farski (approved):
I think we should consider this change temporary until we know it's necessary, given that it's a cost increase.

svevang:
Looks like things are running fine in staging without the extra headroom -- just needed ~2 gigs of RAM to run the allocation processes.

PR #726, opened 2023-11-14
Allow feeder to delete S3 objects

cavis:
Do this or https://github.com/PRX/feeder.prx.org/pull/918

### DIFF
--- a/spire/templates/apps/feeder.yml
+++ b/spire/templates/apps/feeder.yml
@@ -988,6 +988,7 @@ Resources:
               - s3:ListMultipartUploadParts
               - s3:PutObject
               - s3:PutObjectAcl
+              - s3:DeleteObject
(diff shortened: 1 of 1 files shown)


spire/templates/apps/feeder.yml
```diff
@@ -983,6 +983,7 @@ Resources:
             Resource: !GetAtt FeederUploadsBucket.Arn
           - Action:
               - s3:AbortMultipartUpload
+              - s3:DeleteObject
```
cavis, line 986: Would this also need `s3:DeleteObjectVersion`, for `s3.delete_object` calls?

farski: Yes, for a versioned bucket I think you need `DeleteObjectVersion` to do any deleting

PR #729, opened 2023-11-30
Enable boostr impression export

cavis:


spire/templates/apps/dovetail-metrics-export.yml
```diff
@@ -239,3 +239,36 @@ Resources:
...
+      State: ENABLED
+      Targets:
+        - Arn: !GetAtt ExportFunction.Arn
+          Id: !Sub "${RootStackName}.Export.BoostrIntegration"
+          Input: |-
+            {
+              "IntegrationIds": [1],
+              "CompressionType": "NONE",
+              "DestinationFormat": "CSV",
+              "ObjectPrefix": "boostr/",
```
cavis, line 259: In my original doc, I said the format would be:

```
prod/soundrise/boostr/2023-06-12T00:00:00.000Z/FILE_SEQ_ID.csv.gz
```

I guess my only question is - will the default format potentially result in multiple files for the same day, or overwrite?  Probably better to just overwrite, for this use?

cavis:
Tested and working.  Will check in on it after it runs.

PR #730, opened 2023-12-04
Simpler path/file format for boostr exports

cavis:
Requires https://github.com/PRX/dovetail-metrics-export/pull/13

### DIFF
--- a/spire/templates/apps/dovetail-metrics-export.yml
+++ b/spire/templates/apps/dovetail-metrics-export.yml
@@ -257,7 +257,6 @@ Resources:
               "IntegrationsIDs": [1],
               "CompressionType": "NONE",
               "DestinationFormat": "CSV",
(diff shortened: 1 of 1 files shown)


spire/templates/apps/dovetail-metrics-export.yml
```diff
@@ -257,11 +257,12 @@ Resources:
               "IntegrationsIDs": [1],
               "CompressionType": "NONE",
               "DestinationFormat": "CSV",
-              "ObjectPrefix": "boostr/",
+              "ObjectPrefix": "",
```
farski, line 260: I think it'd be better to not include `ObjectPrefix` than have it be an empty string

farski (changes requested):

farski (approved):

PR #742, opened 2024-01-29
CNE data export

cavis:


spire/templates/apps/dovetail-metrics-export.yml
```diff
@@ -244,6 +244,53 @@ Resources:
...
+    Properties:
+      Description: CNE network data export
+      RoleArn: !GetAtt RulesRole.Arn
+      ScheduleExpression: cron(15 0 * * ? *) # 00:15 UTC every day
+      State: ENABLED
+      Targets:
+        - Arn: !GetAtt ExportFunction.Arn
+          Id: !Sub "${RootStackName}.Export.CNENetwork"
+          Input: |-
+            {
```
farski, line 259: I would add an `ObjectPrefix`, just in case we ever want to send different data to the same bucket

farski (changes requested):

farski (approved):

PR #743, opened 2024-02-05
Use native ECS cxontainer instance draining

farski:
Closes https://github.com/PRX/internal/issues/1077

I believe this is essentially a one-to-one replacement for the lifecycle hook Lambda function, but some additional details:

### [Overview](https://aws.amazon.com/blogs/containers/amazon-ecs-enables-easier-ec2-capacity-management-with-managed-instance-draining/)

>No matter how you choose to refresh your EC2 instances, it will cause Amazon EC2 instances that are part of an Auto Scaling Group to be replaced. If an Amazon EC2 instance is stopped and replaced while it is running tasks, then those tasks will be stopped as well. Amazon ECS will detect the loss of any task that is part of an Amazon ECS service. In order to maintain the service’s desired count of tasks, Amazon ECS will launch a replacement task onto a different Amazon EC2 instance in the cluster. But this is a reactive fail-safe that only happens after a container has already been stopped, and the service’s running task count has already dropped below it’s desired count of tasks.

> Managed instance draining is proactive, rather than reactive. Whenever an Amazon EC2 instance is set to be terminated by the Auto Scaling Group, Amazon ECS will temporarily delay the instance termination, and automatically place the instance into draining mode. This draining mode prevents any more task launches on the instance, and causes any service launched tasks running on the instance to be proactively replaced onto new hosts. The task replacement attempts to launch new replacement tasks prior to stopping existing tasks that are currently running on the draining Amazon EC2 instance. You can read more about draining behavior under [Container Instance Draining](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container-instance-draining.html) in the official Amazon ECS documentation.

---

Full docs: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container-instance-draining.html


cavis (approved):
👍 seems like it should work.

Should we manually try this in staging first?  Or I guess we'd just revert this if something goes wrong in staging.

PR #747, opened 2024-02-16
Handle redirect headers on S3 feed objects

farski:
Makes a HEAD request to the same file as the origin request, looks for a `x-amz-meta-dovetail-feed-cdn-redirect` header, if present returns a 301 to that header's value.

- Example without metadata set on S3 object: [`https://staging.publicfeeds.net/f/100/feed-rss.xml`](https://staging.publicfeeds.net/f/100/feed-rss.xml)
- Example with metadata set (`https://example.com`) on object: [`https://staging.publicfeeds.net/f/101/feed-rss.xml`](https://staging.publicfeeds.net/f/101/feed-rss.xml)

Closes https://github.com/PRX/internal/issues/1086

### DIFF
--- a/cdn/publicfeeds-cdn.yml
+++ b/cdn/publicfeeds-cdn.yml
@@ -70,21 +70,45 @@ Resources:
           exports.handler = (event, context, callback) => {
             const request = event.Records[0].cf.request;
 
-            if (request.uri.startsWith('/f/3954/feed-rss.xml')) {
-              callback(null, { status: 301, headers: { location: [{ key: 'Location', value: 'https://feeds.megaphone.fm/technically-optimistic' }] } });
-            } else if (request.uri.startsWith('/f/')) {
+            // The `/f` path is reserved to indicate an f.prxu.org feed origin. The `/f`
+            // path part must be removed to make a valid origin request.
+            if (request.uri.startsWith('/f/')) {
               request.uri = request.uri.slice(2);
+
+              const headUrl = `https://${request.origin.custom.domainName}${request.uri}`;
+
+              // Check the feed file for a redirect marker
+              fetch(headUrl, { method: 'HEAD' }).then((res) => {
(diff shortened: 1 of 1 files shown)


cdn/publicfeeds-cdn.yml
```diff
@@ -70,30 +70,54 @@ Resources:
...
+              // Check the feed file for a redirect marker
+              fetch(headUrl, { method: 'HEAD' }).then((res) => {
+                if (res.headers.has('x-amz-meta-dovetail-feed-cdn-redirect')) {
+                  // The feed file in S3 being requested has been marked with a
+                  // redirect header. Respond with a 301 to the given URL, rather than
+                  // with the actual feed.
+                  callback(null, {
+                    status: 301,
+                    headers: {
+                      location: [
```
cavis, line 89: Does this need a `Cache-Control` header on it?  Not for the client really (301s are permanent)... but for the CDN?  I'm not sure either publicfeeds or f.prxu.org have a default TTL.  So they may just cache it for a real-long-time if you don't provide a max age.

cdn/publicfeeds-cdn.yml
```diff
@@ -70,30 +70,54 @@ Resources:
...
-            } else if (request.uri.startsWith('/f/')) {
+            // The `/f` path is reserved to indicate an f.prxu.org feed origin. The `/f`
+            // path part must be removed to make a valid origin request.
+            if (request.uri.startsWith('/f/')) {
               request.uri = request.uri.slice(2);
+
+              const headUrl = `https://${request.origin.custom.domainName}${request.uri}`;
+
+              // Check the feed file for a redirect marker
+              fetch(headUrl, { method: 'HEAD' }).then((res) => {
```
cavis, line 81: Too cool for async/await, huh?

cavis (approved):
One question on default cache headers, but otherwise looks good to me.

PR #766, opened 2024-05-15
Force FlowLog resource replacement

farski:
To support [the newest](https://aws.amazon.com/blogs/networking-and-content-delivery/introducing-vpc-flow-logs-for-elastic-container-services/) version of flow log format that include ECS data, I think we need to create a new flow log.

This replaces the existing flow log with a new one by changing the resource name, and also update some of the resource definition to the more modern format.


cavis (approved):
👍

PR #767, opened 2024-05-15
Remove Publish stack

farski:
For when we're ready

- [x] This removes the Publish hostname envar from ID, so those references need to be removed from ID before this can merge


cavis (commented):
This looks good to me, but let's maybe hold off until next week?

Also - we should unsubscribe Feeder from CMS "announcements" next week.

PR #778, opened 2024-08-15
Add optional sort key as part of primary key

kookster:


db/dynamodb-global-table.yml
```diff
@@ -194,12 +211,22 @@ Resources:
     Type: AWS::DynamoDB::GlobalTable
     Properties:
       AttributeDefinitions:
-        - AttributeName: !Ref PrimaryKey
-          AttributeType: !FindInMap [DataTypesMap, !Ref PrimaryType, Symbol]
+        - AttributeName: !Ref PartitionKey
+          AttributeType: !FindInMap [DataTypesMap, !Ref PartitionType, Symbol]
+        - !If
```
farski, line 216: See [item #2](https://github.com/PRX/Infrastructure/blob/main/CONTRIBUTING.md#function-syntax)

db/dynamodb-global-table.yml
```diff
@@ -91,12 +97,22 @@ Parameters:
   TableName:
     Type: String
     Description: The unique DynamoDB table name.
-  PrimaryKey:
+  PartitionKey:
```
farski, line 100: The `PrimaryKey` parameter was a required parameter prior to this change, which means that if a stack deployed from this template were to be updated for some reason unrelated to this change, it would lose that parameter. At a minimum I would say that should be documented in the file, but to maintain compatibility it may be worth keeping the parameter name the same, and using the `ParameterLabels` and `Description` to denote that this value is actually a partition key, even if the name remains `PrimaryKey`.

db/dynamodb-global-table.yml
```diff
@@ -91,12 +97,22 @@ Parameters:
...
     Type: String
-    Description: The name of the primary key field
-  PrimaryType:
+    Description: The name of the partition key field
+  PartitionType:
```
farski, line 103: Same as above

db/dynamodb-global-table.yml
kookster: nice! will change `!If` -> `Fn::If:`

db/dynamodb-global-table.yml
kookster: I moved everything for that param back to how it was, trying to make the most minimal change

db/dynamodb-global-table.yml
kookster: reverted!

cavis (approved):
👍 i'm never 100% confident where `AWS::NoValue` will work out.  But this looks good to me - worth trying out.

cavis:
Oh, but `prettier` linting did fail.  (I think the `cfn-lint` stuff is all ignorable right now).

kookster:
> Oh, but `prettier` linting did fail. (I think the `cfn-lint` stuff is all ignorable right now).

apparently it didn't like the amount of indentation - fixed.

db/dynamodb-global-table.yml
```diff
@@ -50,6 +52,10 @@ Metadata:
         default: Primary key field
```
farski, line 52: Could be changed to "Partition key field"

db/dynamodb-global-table.yml
```diff
@@ -50,6 +52,10 @@ Metadata:
...
       PrimaryType:
         default: Primary key type
```
farski, line 54: Could be changed to "Partition key type"

farski (approved):

db/dynamodb-global-table.yml
kookster: done

db/dynamodb-global-table.yml
kookster: done

kookster:
This worked in staging 

https://us-east-1.console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks/stackinfo?filteringStatus=active&filteringText=&stackId=arn%3Aaws%3Acloudformation%3Aus-east-1%3A424798965986%3Astack%2Fstaging-dovetail-listener-frequency-ddb-global%2F5487fef0-76bd-11ef-9472-0e45891c840f&viewNested=true

I'm going to call it done, I'll make any necessary changes that might come up in testing in a new PR / commits

PR #779, opened 2024-09-10
Treat missing data for some alarms as missing

farski:
Closes https://github.com/PRX/internal/issues/1131

To be seen how often these don't have data. I'm assuming only during fairly major AWS issues. But we can dial things back if it's noisy. I went with `missing` rather than `breaching` so at least make it obvious when we see alarms because of this change.


cavis (approved):
👍 works for me

PR #783, opened 2024-09-17
Upgrade Router scaler function to Node.js 20 and AWS SDK v3

### LINKED ISSUE #733: Upgrade Dovetail Router feed scaler Lambda function to Node.js 20

farski:
This probably needs at least one real-world test run. Some of the v3 SDK commands work a little differently than v2. I think I accounted for all of them, but worth checking.

Closes #733

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "@aws-sdk/client-cloudwatch": "3.x",
     "@aws-sdk/client-codebuild": "3.x",
     "@aws-sdk/client-codepipeline": "3.x",
+    "@aws-sdk/client-dynamodb": "3.x",
     "@aws-sdk/client-ec2": "3.x",
     "@aws-sdk/client-ecs": "3.x",
     "@aws-sdk/client-eventbridge": "3.x",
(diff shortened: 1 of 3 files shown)


spire/templates/apps/dovetail-router.yml
```diff
@@ -940,15 +953,18 @@ Resources:
         };
 
         exports.handler = async (event) => {
+          console.log(JSON.stringify(event));
```
cavis, line 956: Temporary for the "real world test run"?

package.json
```diff
@@ -16,6 +16,7 @@
     "@aws-sdk/client-cloudwatch": "3.x",
     "@aws-sdk/client-codebuild": "3.x",
     "@aws-sdk/client-codepipeline": "3.x",
+    "@aws-sdk/client-dynamodb": "3.x",
```
cavis, line 19: So what is this package.json used for?  Are you running the code for these functions locally somehow?

cavis (approved):
👍 works for me!

package.json
farski: Really just so that my IDE has type info

spire/templates/apps/dovetail-router.yml
farski: I like logging events in Lambdas so that if I ever need to run them manually there are good samples around. I can remove though, if you want.

PR #784, opened 2024-09-22
Template updates for Frequency Capping

kookster:
- [x] Analytics Lambda
- [x] Dovetail Router


spire/templates/apps/dovetail-router.yml
```diff
@@ -739,6 +741,25 @@ Resources:
             Effect: Allow
             Principal:
               Service: ecs-tasks.amazonaws.com
+          - Action:
+              - dynamodb:BatchGetItem
+              - dynamodb:ConditionCheck
+              - dynamodb:DescribeTable
+              - dynamodb:DescribeTimeToLive
+              - dynamodb:GetItem
+              - dynamodb:Query
```
kookster, line 750: DTR should only need to read data, no write/update/delete actions

spire/templates/apps/dovetail-router.yml
```diff
@@ -739,6 +741,25 @@ Resources:
...
+            Effect: Allow
+            Resource: !Split
```
farski, line 752: This split/sub/join/split insanity was only needed for a short time when the arrangements table was being moved from a regional table to a global table, and we were accessing both. I don't believe that's the case any more.

So for this I would either simplify this back to a single `!Ref`, or probably `FrequencyDynamodbTableName` should to be renamed to something that indicates that it can/does contain multiple table names

spire/templates/apps/dovetail-analytics.yml
```diff
@@ -722,6 +725,285 @@ Resources:
...
+        - { Key: prx:cloudformation:root-stack-name, Value: !Ref RootStackName }
+        - { Key: prx:cloudformation:root-stack-id, Value: !Ref RootStackId }
+        - { Key: prx:ops:environment, Value: !Ref EnvironmentType }
+        - { Key: prx:ops:cloudwatch-log-group-name, Value: !Ref AnalyticsFrequencyFunctionLogGroup }
+        - { Key: prx:dev:family, Value: Dovetail }
+        - { Key: prx:dev:application, Value: Analytics }
+      Threshold: 0
+      TreatMissingData: notBreaching
+
+  AnalyticsFrequencyFunctionInsertsMetricFilter:
```
farski, line 973: The metrics generated by these next few metric filters don't seem to be used anywhere. Do we want to put them on a dashboard or alarm on them? Or are they just for reference purposes, should the need arise?

spire/templates/apps/dovetail-analytics.yml
```diff
@@ -722,6 +725,285 @@ Resources:
...
+        - { Key: prx:dev:family, Value: Dovetail }
+        - { Key: prx:dev:application, Value: Analytics }
+  AnalyticsFrequencyFunctionElevatedErrorAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: IsProduction
+    Properties:
+      AlarmName: !Sub WARN [Dovetail-Analytics] Frequency Lambda function <${EnvironmentTypeAbbreviation}> INVOCATIONS ERRORS (${RootStackName})
+      AlarmDescription: !Sub >-
+        ${EnvironmentType} Dovetail Analytics Frequency Lambda function is
+        failing, but tktktk.
```
farski, line 819: tktktk

spire/templates/apps/dovetail-analytics.yml
```diff
@@ -722,6 +725,285 @@ Resources:
...
+                - dynamodb:DeleteItem
+                - dynamodb:DescribeTable
+                - dynamodb:DescribeTimeToLive
+                - dynamodb:GetItem
+                - dynamodb:PutItem
+                - dynamodb:Query
+                - dynamodb:UpdateItem
+              Effect: Allow
+              # TODO: can this be done with an AWS::Partition Sub?
+              Resource: !Split
```
farski, line 772: See other comment about this split/sub/join stuff

spire/templates/apps-300A.yml
```diff
@@ -152,6 +152,8 @@ Resources:
         DovetailCountedKinesisStreamName: !Ref DovetailCountedKinesisStreamName
         DovetailRouterHosts: !Sub /prx/${EnvironmentTypeAbbreviation}/dovetail-analytics/DOVETAIL_ROUTER_HOSTS
         DovetailRouterApiTokens: !Sub /prx/${EnvironmentTypeAbbreviation}/dovetail-analytics/DOVETAIL_ROUTER_API_TOKENS
+        FrequencyDynamodbTableName: !Sub /prx/${EnvironmentTypeAbbreviation}/dovetail-analytics/FREQUENCY_DDB_TABLE
```
farski, line 155: These Parameter Store parameters are using a naming convention that predates the [current standard](https://github.com/PRX/internal/wiki/AWS:-Parameter-Store#spire), and the few parameters that remain from back then are [slated to be cleaned up](https://github.com/PRX/internal/issues/887#issuecomment-1538922226).

It would be ideal if we didn't add any new parameters that don't follow the current standard, but if it makes more sense to deal with this later, let's make sure they at least get added to that issue.

spire/templates/apps/dovetail-router.yml
```diff
@@ -739,6 +741,25 @@ Resources:
...
+              - ","
+              - Fn::Sub:
+                  - arn:aws:dynamodb:*:*:table/${inner}
+                  - inner:
+                      Fn::Join:
+                        - ",arn:aws:dynamodb:*:*:table/"
+                        - !Split [",", !Ref FrequencyDynamodbTableName]
+          - Action: sts:AssumeRole
+            Effect: Allow
+            Resource: !Split [",", !Ref FrequencyDynamodbAccessRoleArn]
```
farski, line 762: Similar situation here: I believe we expect this to be a single role ARN now, so no need to `Split`.

spire/templates/apps/dovetail-router.yml
```diff
@@ -739,6 +741,25 @@ Resources:
             Effect: Allow
             Principal:
               Service: ecs-tasks.amazonaws.com
+          - Action:
```
farski, line 744: This statement needs to go under `Policies`, not the `AssumePolicyDocument`. See https://github.com/PRX/Infrastructure/blob/main/spire/templates/apps/feeder.yml#L599 for an example

farski (changes requested):

spire/templates/apps/dovetail-analytics.yml
kookster: heh, copied from other alarms in the same file with the same `tktktk` description

spire/templates/apps/dovetail-analytics.yml
kookster: Looking at the other lambdas in this template, I created equivalent metrics to what they had defined.
Is there someplace else I need to add these, some other dashboard template I don't remember?

farski (approved):

PR #785, opened 2024-09-24
Update Castle DB version

farski:
`13.13` is the version already being used by both stag and prod Castle DBs, the template seems to have gotten out of sync


cavis (approved):
👍

PR #789, opened 2024-11-01
Fix which stream of events we use for freq cap counting

kookster:
wrong stream, need the output of the counts, not the input


cavis (approved):
That'll do it!

PR #797, opened 2025-01-06
Add instance-type placement constraint to ECS tasks

farski:
This adds `PlacementConstraints` to each existing ECS service, ensuring that associated tasks are only placed on `x86_64` instances.

This shouldn't make any functional change to existing deployments, since all instances are `x86_64`.


cavis (approved):
👍 looks good to me.

Any way to confirm this in staging?  Force-redeploy a service or something?

farski:
@cavis Yeah, will do some clicking to try to make sure it works

PR #799, opened 2025-01-13
Shared ALB WAF

farski:
Adds a WAF to the shared ALB. Does not change the Dovetail Router ALB.

Meant to mimic the setup/rule that we currently have deployed.

This shouldn't be deployed until the existing WAF is torn down.

### DIFF
--- a/spire/templates/shared-alb.yml
+++ b/spire/templates/shared-alb.yml
@@ -442,6 +442,20 @@ Resources:
         Allow: {}
       Description: !Sub WAF for Spire ${EnvironmentType} shared ALB
       Scope: REGIONAL
+      Rules:
+        - Action:
+            Challenge: {}
+          Name: reject-exchange-geo
+          Priority: 20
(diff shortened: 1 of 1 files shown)


spire/templates/shared-alb.yml
```diff
@@ -435,30 +435,44 @@ Resources:
...
-  #     VisibilityConfig:
-  #       CloudWatchMetricsEnabled: false
-  #       MetricName: !Sub ${Alb.LoadBalancerName}-WAF
-  #       SampledRequestsEnabled: false
-  # WafAssociation:
-  #   Type: AWS::WAFv2::WebACLAssociation
-  #   Properties:
-  #     ResourceArn: !Ref Alb
-  #     WebACLArn: !GetAtt Waf.Arn
+  Waf:
```
farski, line 438: Do we want to put this in its own template, so we can more easily share it between this ALB and the DT ALB?

PR #803, opened 2025-01-24
Service capacity providers

farski:
- Reverts the default capacity provider strategy for the ECS cluster to use only the x86 capacity provider (this only affects ECS services (not tasks) that are created from this point forward, it's essentially a no-op)
- Intended to explicitly set the capacity provider strategy on every ECS service and non-service task (things that use `runTask`, like cron workers and Augury slow worker Lambda function).

If this PR misses any services/tasks, that's not really a big deal (at this point), since they will continue to work as they have been.

The bigger deal would be if this gets the provider strategy _wrong_ (aarch64 when it should be x86, a malformed definition, etc) for any of the resources that it is affecting. This is redefining the capacity strategy for all ECS services, some which haven't been touched in years. The intention, though, is that the new explicit definition is identical to the existing configuration. CloudFormation will actually touch all of them, though, since it's new config in the CFN resources.

Key things to look for:
- That the `!GetAtt`s in `root.yml` are getting the correct value from the correct ASG template (no crossed wires where an aarch64 name ends up in an x64 parameter)
- That each service/task is using the correct provider for the desire platform (everything except Insights and ID should use x86)

This is probably not a Friday deploy.

### DIFF
--- a/spire/templates/apps/augury.yml
+++ b/spire/templates/apps/augury.yml
@@ -1180,13 +1180,6 @@ Resources:
         // ECS runTask parameters
         const runParams = (name) => {
           return {
-            capacityProviderStrategy: [
-              {
-                capacityProvider: process.env.X86_64_ASG_CAPACITY_PROVIDER_NAME,
-                base: 0,
-                weight: 1
-              }
-            ],
             cluster: process.env.ECS_CLUSTER_ARN,
             launchType: 'FARGATE',
             networkConfiguration: {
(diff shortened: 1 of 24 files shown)


spire/templates/apps/augury.yml
```diff
@@ -1170,6 +1180,13 @@ Resources:
         // ECS runTask parameters
         const runParams = (name) => {
           return {
+            capacityProviderStrategy: [
```
cavis, line 1183: So is the idea that... when we move Augury over to arm, we'll need the Fargates to also launch using that capacity provider?

cavis (approved):
👍 looks good to me

spire/templates/apps/augury.yml
farski: I think we'd actually want to set [this runtime config](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-runtimeplatform.html) on the task def. I forgot how Fargate worked.

PR #805, opened 2025-02-06
Use global endpoint parameters for Aurora DBs

farski:
Part of https://github.com/PRX/internal/issues/1185 but needs some cleanup work once this is deployed


cavis (approved):
👍

PR #807, opened 2025-04-09
Remove dupe alarm

svevang:
This is already handled by e.g. `WebLoggedErrorsAlarm` alarm using [the metrics filter on the log level](https://github.com/PRX/Infrastructure/blob/fbfd72cb3dc76f8cadf68d5731efd0689a169d0c/spire/templates/apps/augury.yml#L673-L682)


cavis (approved):
👍

PR #812, opened 2025-05-23
Use hosted Valkey for apps, switching to ReplicationGroup

farski:
Should essentially be recreating [what we had](https://github.com/PRX/Infrastructure/blob/f7dd8140a90f328c9c47e301e98857cd1f1cb341/spire/templates/shared-app-redis.yml#L77), but using `AWS::ElastiCache::ReplicationGroup` instead of `AWS::ElastiCache::CacheCluster`, which doesn't support valkey.


cavis (approved):
Does this change what redis mode we're running in?  (Clustered vs non-clustered).

Hopefully not a problem, as I don't think our apps connect to a specific db number. 

👍 looks good to me.

PR #816, opened 2025-06-17
Remove Castle DB instance in secondary regions

farski:
- Make the DB instance conditional, which should tear down the existing instance in us-west-2
- Returns bogus placeholder values for address, port, and instance ID when the instance doesn't exist

The two things that currently depend on those values are:
- Augury task definitions use them as environment variables. Since we don't run Augury tasks in secondary regions, the bogus values should never cause a problem
- Jump servers also use them as envars (`export PRX_CASTLE_POSTGRES_HOST=${CastlePostgresInstanceEndpointAddress}`). I think the bogus values would only cause a problem if someone tried to connect to the Castle DB in a region that we don't use, so I'm not worried about that. There is probably a cleaner way to handle that, and exclude those values from the jump server ENV entirely if we really wanted

### DIFF
--- a/spire/templates/apps/castle.yml
+++ b/spire/templates/apps/castle.yml
@@ -43,6 +43,7 @@ Parameters:
 Conditions:
   IsProduction: !Equals [!Ref EnvironmentType, Production]
   IsPrimaryRegion: !Equals [!Ref RegionMode, Primary]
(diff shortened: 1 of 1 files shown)


cavis (approved):
👍

PR #820, opened 2025-07-10
Auth Proxy fixes

cavis:
Saw a couple requests this morning that exceeded the 192mb memory this lambda had.  No idea which, as we don't log that.

Bumping the memory a bit and logging the feed urls for future debugging.

### DIFF
--- a/spire/templates/apps/feeder.yml
+++ b/spire/templates/apps/feeder.yml
@@ -1226,6 +1226,8 @@ Resources:
             return text(403, 'nope, no feed for you!', 300);
           }
 
+          console.log('requesting', `http://${process.env.PUBLIC_FEED_DOMAIN}${event.path}`);
+
           const resp = await new Promise((resolve) => {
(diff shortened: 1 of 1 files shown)


spire/templates/apps/feeder.yml
```diff
@@ -1226,6 +1226,8 @@ Resources:
             return text(403, 'nope, no feed for you!', 300);
           }
 
+          console.log('requesting', `http://${process.env.PUBLIC_FEED_DOMAIN}${event.path}`);
```
kookster, line 1229: +1 yeah, that's nice to have regardless

spire/templates/apps/feeder.yml
```diff
@@ -1254,7 +1256,7 @@ Resources:
           };
           return { statusCode: 200, body, headers };
         };
-      MemorySize: 192
+      MemorySize: 400
```
kookster, line 1259: good, still not huge - worth it to avoid errors.

kookster (commented):
Approved!

PR #824, opened 2025-07-14
Give feeder tasks access to the uploads bucket

kookster:
The worker for Megaphone imports "simulates" uploading the audio to the uploads bucket, so the task needs access to wrote to this bucket.


farski (approved):
I don't think the ACL permissions are useful anymore (probably never were for us, not sure why they got included), but not doing much harm either.

kookster:
oh I see what you mean - yeah, I'm copying permissions from elsewhere in the doc for consistency, but yeah, we could clean those out.

PR #838, opened 2025-09-19
Reschedule pg-dump cron

svevang:
Moves the scheduled time to run pg_dump from 00:15am UTC to 04:30am UTC.

We have some other transaction heavy workloads for augury that this can work around:
- 00:00 UTC midnight reallocation
- 08:00am Collection flight resmoothing

And in general this pushes the pg_dump invocation further away from regular office hours.

### DIFF
--- a/etc/pgdump-to-s3/template.yml
+++ b/etc/pgdump-to-s3/template.yml
@@ -160,7 +160,7 @@ Resources:
     Properties:
       Description: Dump Augury staging database
       RoleArn: !GetAtt RulesRole.Arn
-      ScheduleExpression: cron(15 0 * * ? *) # 00:15 UTC every day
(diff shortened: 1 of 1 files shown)


cavis (approved):
👍 makes sense ... lot of other cron churn happening right after midnight.

svevang:
I know we don't have a reader instance currently, but a deeper solution may be to run the pg_dump on a postgres reader node.  The Aurora readers are sitting behind physical replication that is tied to the MVCC machinery. So no locks on the writer instance, right? We'd only see some delays in vacuuming as the tx id tied to the pg_dump would hang out until the pg_dump finishes. But assuming the pg_dump took less than 24 hours, it should be a blip in the metrics.

PR #844, opened 2025-10-21
Reduce observability settings for The World

farski:
Closes https://github.com/PRX/The-World-CMS-Wordpress/issues/28


kookster (approved):
Makes sense, and saves some do-re-mi

PR #892, opened 2026-04-27
Move New Relic key to secret

farski:
Different apps use different names for this env var, so one thing to look at is make sure I copied the correct one in each file


cavis (approved):
👍 looks good to me - the ENV names all seem to match.

PR #894, opened 2026-04-28
Remove Styleguide

farski:
Also need to remove `/prx/${EnvironmentTypeAbbreviation}/Spire/Styleguide/pkg/docker-image-tag` from SSM, and I guess images from ECR


cavis (approved):
👍 we barely knew thee

PR #900, opened 2026-05-15
Remove Spire instance SG SSH rule

farski:
We may want to make sure we aren't forgetting about any non-human SSH connections we might be making to these instances


cavis (approved):
Hopefully nothing is still using ssh?  Guess we'll find out.

PR #903, opened 2026-05-27
Set LRU memory policy on shared app Valkey

farski:
Also removes a wayward tag

Closes https://github.com/PRX/internal/issues/1475


cavis (approved):
👍

PR #913, opened 2026-08-03
Update media:content as well for auth codes

### LINKED ISSUE #912: Also add auth code to media:content, not just enclosure, in the private feeds proxy
https://github.com/PRX/Infrastructure/blob/main/spire/templates/apps/feeder.yml#L1298-L1301

kookster:
fixes #912

### DIFF
--- a/spire/templates/apps/feeder.yml
+++ b/spire/templates/apps/feeder.yml
@@ -1298,7 +1298,11 @@ Resources:
           const body = data.replace(
             /<enclosure url="([^"]+)"/g,
             `<enclosure url="$1?auth=${auth}"`,
+          ).replace(
+            /<media:content fileSize="([^"]+)" type="([^"]+)" url="([^"]+)"/g,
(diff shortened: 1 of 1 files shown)


spire/templates/apps/feeder.yml
```diff
@@ -1298,7 +1298,11 @@ Resources:
           const body = data.replace(
             /<enclosure url="([^"]+)"/g,
             `<enclosure url="$1?auth=${auth}"`,
+          ).replace(
+            /<media:content fileSize="([^"]+)" type="([^"]+)" url="([^"]+)"/g,
```
cavis, line 1302: Just double checked prod ... and `enclosure_file_size` / `enclosure_content_type` are never nil/blank for any published ep.  So this should work.

cavis (approved):
👍 looks good to me

spire/templates/apps/feeder.yml
kookster: oh good call, I didn't think about that case, ty!

PR #917, opened 2026-09-10
Domain validation for supporting cast using fastly

kookster:
adding in what was provided:
```
Type: CNAME
Name: _acme-challenge.subscribe.prx.org
Value: h9ic2fdq3umr9pvk2v.fastly-validations.com
```


dns/prx.org-hosted_zone.yml
```diff
@@ -1094,6 +1094,17 @@ Resources:
             - c4j2.domainkey.u34190588.wl089.sendgrid.net.
           TTL: "3600"
           Type: CNAME
+  SupportingCast:
+    Type: AWS::Route53::RecordSetGroup
+    Properties:
+      Comment: Supporting Cast
+      HostedZoneId: !Ref HostedZone
+      RecordSets:
+        - Name: !Sub _acme-challenge.subscribe.${Domain}
```
farski, line 1103: This record is not SupportingCast-specific, so it shouldn't go in a `SupportingCast` record set group. It belongs I guess in an ACME record set group.

(The reason being, if we ever started using `subscribe.prx.org` as the domain for something not related to SupportingCast, but which still required ACME verification, we would not want this record to continue to live under a SupportingCast group.)

farski (changes requested):

